PR #9, opened 2020-01-30
Added final common functionalities for CLSpectrum class

jordiferrero:
The following common simple functionalities have been added:

- Background subtraction
- Background storage as metadata (upon loading file from folder for Attolight)
- Cosmic rays removal (spike removal). Very simple, needs further modifications.

### DIFF
--- a/lumispy/signals/cl.py
+++ b/lumispy/signals/cl.py
@@ -107,7 +107,7 @@ class CLSpectrum(Signal1D):
 
         Parameters
         ---------------
-        background : array[wavelength, bkg]
+        background : numpy.array[wavelength, bkg]
             OPTIONAL: Bakground array with two columns: [wavelenght, bkg]. Length of array must match signal_axes size.
 
         inplace : boolean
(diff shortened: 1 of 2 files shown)


jlaehne:
What is the advantage of your cosmic rays removal over the hyperspy `spikes_removal_tool`? Maybe add a comment on that also to the docstring.

jordiferrero:
As far as I am aware, the `spikes_removal_tool` in Hyperspy relies on a GUI and requires manual inspection and manual tweaking of parameters. 
The function written here is simpler and can be run quickly. However, I did intend to integrate the Hyperspy functionality too, because it is more complete and better. I did not find the time yet to do so.

jlaehne:
Well, no need to duplicate the hyperspy functionality here, I think (in case it would evolve, we would always have to adapt). As far as I understand, that is the idea of creating an hyperspy extension that we should be able to call their functions as well.

But it would be good to add details to the docstring of your functions on how it identifies the spikes and add a note that there is also the hyperspy function as alternative.

PR #32, opened 2020-05-04
Add Function: Calibrating WL axis using MeanSpec

arminbarthel:
Add function for calibrating wavelength signal axis using Mean spectrum file produced by Attomap software.

### DIFF
--- a/lumispy/io_plugins/attolight.py
+++ b/lumispy/io_plugins/attolight.py
@@ -216,6 +216,29 @@ def file_reader(filename, *args, **kwds):
 
         return cl_object
 
+    def _calibrate_signal_axis_wavelength_meanspec(filename, cl_object):
+        ms_file = "HYPCard - Mean spectrum.txt"  # how attomap software names mean spectrum file
+        file = os.path.join(filename, ms_file)  # filename is str; the absolute folder path where the ms_file_name exists
+
+        binning = cl_object.Acquisition_instrument.CCD.binning
+        WL = []  # create empty wavelength array
+
+        # Read in wavelengths from Mean spectrum file, which are in the first column.
+        with open(file, 'r') as f:
+            next(f)
+            for line in f:
+                sep = line.split()
+                WL.append(float(sep[0]) * 1e6)  # wavelength given in mm in file. Convert to nm.
+
+        dx = cl_object.axes_manager.signal_axes[0]  # want to attribute modification of wavelength axis to cl_object
+
+        dx.name = 'wavelength'
+        dx.scale = (WL[len(WL) - 1] - WL[0]) / len(WL) * binning  # Wavelength increment
+        dx.offset = WL[0]  # Wavelength offset
+        dx.units = '$nm$'
+
+        return cl_object
+
     def _calibrate_navigation_axis(cl_object):
         # Edit the navigation axes
         x = cl_object.axes_manager.navigation_axes[0]


jordiferrero:
Hi Armin,
Thanks for helping build new functionalities in Lumispy! 
There is a main problem with the function you have written: it is not accessible at any point and it is not called at any point.

I am not sure what the best strategy to fix this would be:
- Add an extra parameter in `lum.load()` called something like (`calibrate_signal_axis_from_meanspec_file`) so if this parameter it's true, it runs your function. If this parameter is not true, it will run the normal `calibrate_signal_axis` function. All this can be easily added in this same file you've edited.
- Maybe we should have this function as a class function of the"LumiSpectrum" class, so anyone can calibrate the signal axis using an external file. This would mean you would have to generalise this function to any `.txt` file. Then in this function could check in the metadata if the `acquisition_system` is the `cambridge_attolight` and if true, run the very specific processes you have written down.

I personally think the second is more useful, but happy to discuss.

lumispy/io_plugins/attolight.py
```diff
@@ -216,6 +216,29 @@ def _calibrate_signal_axis_wavelength(cl_object):
...
+
+        binning = cl_object.Acquisition_instrument.CCD.binning
+        WL = []  # create empty wavelength array
+
+        # Read in wavelengths from Mean spectrum file, which are in the first column.
+        with open(file, 'r') as f:
+            next(f)
+            for line in f:
+                sep = line.split()
+                WL.append(float(sep[0]) * 1e6)  # wavelength given in mm in file. Convert to nm.
```
jordiferrero, line 231: Also, maybe using the `numpy.loadtxt()` function here is faster. Check how to [here](https://numpy.org/doc/stable/reference/generated/numpy.loadtxt.html?highlight=loadtxt#numpy.loadtxt).

jlaehne:
As I do not know the Attolight system, can you describe why you have to do this correction of the calibration and exactly what the procedure is doing? Then, I could maybe give feedback whether the function is of more general interest.

jordiferrero:
The attolight system has given us (in Cambridge) many headaches in terms of the wavelength calibration. The system does not save the wavelength axes, so we had to find different ways of calibrating it manually.
One way is to export the mean spectrum over a hyperspectral map (the input file of this functionality here) and then use the x-axis of that (so the wavelength) as the calibrated wavelength array for the signal object.

I guess the question is whether this functionality can be extended to a function that can easily change the signal axis (by default a linear axis) to a desired array of interest (no need to be linear) easily from an external calibration file (e.g. the mean spectrum file for the Attolight system).
I hope that makes sense.

If you think this functionality is useless for any signal type, then we can simply include it in the attolight loading plugin... but I personally think that changing the signal axis in hyperspy has always been a bit tedious for beginners (you need to invoke the `axes_manager`...)

gkusch:
There are two underlying issues here.
1) The way the system Export data, the Attolight system has three ways of exporting/saving data.
     - Proprietary Format, can only be read by their own Analysis Software, whic is super crummy and can't even do basic fitting. The proprietary export has the Wavelength axis in it as well as all the metadata (as far as I know)
   - Binary File, the binary file only contains all the intensity information (intensity on each detector channel in each pixel) does not contain any other information (metadata, wavelength axis)
- Textfile (I think ascii format) the text file would contain intensity and wavelength information but has all the problems inherent to textfiles, the most drastic one being that the system crashes tryiing to write a 512x512 map as a textfile, this option is disabled for my own sanity's sake

The way we do it at hte moment is the followiing: write binary and proprietary file, read proprietary file with their software, export mean spectrum for wavelength calibration, read that into hyperspy and then do actual data analysis. This works but is for obvious reasons not ideal. I did try and solve the spectrometer equation for the setup and while I am perfectly capable of recreating the examples given by the horiba website, the values I am given for the spectrometer do not recreate the spectrum and extracted from the proprietary software and thus we are not using it (~3-5nm error)

2) The other problem, necessitating a wavelength correction. The attolight system is a full imaging system, That means that on large field of views (i.e. 300µm) light from either edge of the map, lands on a different spot on the grating. leading to the same wavelength being recorded on different pixels of the detector, while this does have advantages in the optical setup it necessitates a correction at the end to virtually but the intensities on the right calibrated wavelenth. This is curently being done by a script I experimentally derived by measuring bulk GaN on different magnifications and deducting the wavelength shift.

I hope this helps clarifying some bits.

-

jlaehne:
Yep, thanks.

Point 1 seems rather Attolight specific (choice of mean spectrum being arbitray I guess, you could also just export any single spectrum from the map). Sounds to me like it would make sense to have it as part of the io-routine as long as attolight does not provide a proper export.

The use case for calibrations that I would have is that I realize only afterwards that my calibration (done on the setup with a reference lamp or if I want to improve further on a sample with a sharp emission line in the right spectral range) was a bit off and then I would basically shift the axis by a few pixels to correct. So far I do not use any specific function for that, but I guess it would be a fairly common use case.

Point 2 is a broader problem, not only for the attolight. Though having a field of view of 1 order less in the MonoCL4, The effect is not very pronounced. But I have also observed such shifts in emission energy in maps of GaN band edge emission. However, any correction here is pretty instrument specific - depending additionally on wavelength and zoom. So it may be also hard to implement a more general function.

jordiferrero:
Just a clarification of poin 1: Calibration of wavelength is already implemented when you load an attolight .bin file, if a background file from the Attolight system is stored in the folder (which is fairly easy to do during the experiment). That already works well and has been tested.
This functionality would allow to not only be able to calibrate using the background file but also the mean spectrum file, that's all.

jlaehne:
What is the status of this PR in relation with #58 ?

gkusch:
The idea of #58 was to make this one here superflous for the wavelength calibration.

jordiferrero:
I think this can be closed.

PR #30, opened 2020-05-02
Add CI functionalities

jordiferrero:
Added TravisCI (with only 2 Linux machines) and Coveralls.

PR #33, opened 2020-05-06
Continuous Dev part 2

jordiferrero:
Following from #30 and @ericpre comments, some changes and additions are made to integrate CI.


azure-pipelines.yml
```diff
@@ -0,0 +1,74 @@
...
+
+steps:
+- template: azure_pipelines/clone_ci-scripts_repo.yml@templates
+- template: azure_pipelines/activate_conda.yml@templates
+- template: azure_pipelines/setup_anaconda_packages.yml@templates
+
+- bash: |
+    source activate $ENV_NAME
+    # Install package using pip
+    pip install .[tests]
```
ericpre, line 59: I would create the file `conda_environment.yml` and conda_environment_dev.yml` and replace this line with
```
pip install . --no-deps
```
The `--no-deps` will make sure you don't mix pypi and conda packages, which is not advised because there are not ABI compatible and make sure all you dependencies are on conda-forge - even if at the moment, the last point is not really relevant.

jordiferrero: Thank you!
Now this is resolved.

PR #36, opened 2020-05-07
Test using lumispy as endpoint

jordiferrero:
Test if endpoint:lumispy works.


jlaehne:
@jordiferrero is this PR still relevant or could we just close it?

jordiferrero:
We can close it because I never managed to make that work.
So the Azure tests go though my personal azure endpoint usename instead of the lumispy one.

I think it is OK as Azure is working well.

ericpre:
Yes, if this is working fine and as long as this stays like this, no need to take any actions!

Issue #14, opened 2020-04-01
Energy conversion


jlaehne:
My code for energy conversion relies on the non-linear `DataAxis` that is not part of the HyperSpy release yet, but might hopefully become a pre-release soon. Should I still contribute it here already for you to use with the right hyperspy development branch?

It contains some additional features like a wavelength dependent refractive index of air based on an analytical function (therefore it can not be a `FunctionalDataAxis`)

jordiferrero:
I was thinking about this yesterday too.
If we do develop the code now relying on the `non-linear-axes` Hyperspy branch, it is quite tedious to add it as a dependency in the `setup.py` file (because upon installation it would git clone the branch and then install it, which takes longer than just using pip install the released versions).

Therefore I see two options:

1. To create a `dev` branch in lumispy (until Hyperspy adds the functionality)
2. To add it as the `master` branch and clarify in the docstring that it relies on the hyperspy dependency which is in development.

I am happy with both, as lumispy is in a very early stage. Feel free to decide 1 or 2 and contribute!

jlaehne:
I also think both are good options. The second might be a little easier for us to handle so that we don't always have to switch branches in lumispy also (in HyperSpy I basically live in the non-linear-axis branch for more than a year already). I'll try to find some time in the next days to add that.

jordiferrero:
@jlaehne What is the best way to ensure that anyone using LumiSpy works with the non-linear-axis branch? 
I see 2 options:

- Force the installation of this branch during installation (mandated by the `setup.py` file) which I have tried and gives me many issues
- Every time we add a function that relies on non-linear-axis branch, we need to add Error handeling (with a note suggesting the user to install the Hyperspy branch instead of the master version).

For now I think the second is better, as it makes installaton easier.

jlaehne:
Yes, I will add a note to the docstring and will see how to appropriately raise an error if the DataAxis object is not available.

PR #37, opened 2020-05-12
Energy conversion of signal axis

jlaehne:
### Description of the change
As discussed in #14, I implemented and extended my function to convert the signal axis to energy. Includes wavelength dependent permittivity of air (minor correction for very precise data). Corrects intensities so that integrals are correct. Defaults to nm, but can handle µm if explicitly set.
Includes two helper functions `nm2eV` and `eV2nm` for back and forth calculation.

### Progress of the PR
- [X] Raise error when non_uniform_axes functionality is not available from hyperspy,
- [X] Convert to object oriented function,
- [X] Add inplace functionality
- [x] Signal type and metadata for inplace=False
- [x] add tests,
- [x] ready for review.

### DIFF
--- a/lumispy/__init__.py
+++ b/lumispy/__init__.py
@@ -43,6 +43,8 @@ from .signals.luminescence_transient import LazyLumiTransient
 
 from .io import load
 
+from .utils.axes import *
+
 from . import release_info
 
 __version__ = release_info.version
--- a/lumispy/signals/luminescence_spectrum.py
+++ b/lumispy/signals/luminescence_spectrum.py
@@ -39,20 +39,25 @@ class LumiSpectrum(Signal1D, CommonLumi):
         super().__init__(*args, **kwargs)
 
 
-    def to_eV(self,inplace=True):
+    def to_eV(self,inplace=True,jacobian=True):
         """Converts signal axis of 1D signal to non-linear energy axis (eV) 
         using wavelength dependent permittivity of air. Assumes WL in units of 
         nm unless the axis units are specifically set to µm.
         
         The intensity is converted from counts/nm (counts/µm) to counts/meV by 
         doing a Jacobian transformation, see e.g. Wang and Townsend, J. Lumin. 
-        142, 202 (2013).
+        142, 202 (2013), which ensures that integrated signals are correct also
+        in the energy domain.
         
         Input parameters
         ----------------
         inplace : boolean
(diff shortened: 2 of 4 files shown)


jordiferrero:
Thanks @jlaehne for starting this implementation!
It looks good to me! 
I have some comments:
- Using the `.to_eV` function creates a new object with `Signal1D` class (not the luminescence specific classes)
- Using the `.to_eV` function will lose the metadata from the original signal?
Is there a way to get around this issues?

jlaehne:
Good points. So far it is mainly the hack I was using personally.

I could try to also have the creation of the extra signal object optional and otherwise replace the existing object. As the data has to be inverted along the signal axis and the axis replaced, I refrained from messing in place so far. For large signals, it may though be attractive to not create a whole copy. What do you think?

jordiferrero:
I personally prefer the modification of the existing signal by default, because of memory usage mainly and also you run away from problems of losing stored metadata, models... which is annoying.

Maybe you can include the `inplace` parameter (like in the `.map()` function which can trigger the creation of a new object if False.

jlaehne:
Converted it to an object oriented function. For the `luminescence_spectrum` class, so there will need to be a slightly different implementation for time-resolved data, once we add that (because we will have two signal axes and have to convert the right one).
Inplace function should work now, for `inplace=False` I need to use the right Signal type and copy metadata.

jlaehne:
Should all be implemented.

Coverage seems to have some trouble. I get 46% and not 0% locally. Also the Travis fail does not seem related to the new code.

jordiferrero:
I'll have a look at the issues with Travis and coveralls... 
For now I will merge this PR as it is working well. Thank you!
It would be nice to have a demo for that in `lumispy-demos`. I will try to create that.

jordiferrero:
@jlaehne I have also added a section in the installation guide explaining how to install the development hyperspy version of `non-uniform-axes`.

jlaehne:
> It would be nice to have a demo for that in `lumispy-demos`. I will try to create that.
I could do that as well, just wanted to wait for the PR to be merged. Just let me know how you prefer.

I also started working on an equivalent function for wavenumber (to_invcm). For e.g. Raman community that is essential. Only, I do not have so much experience with Raman myself so that I might try to briefly discuss with a colleague before I put it forward.

PR #39, opened 2020-10-04
Remove io_plugins and attolight dependencies

jordiferrero:
Due to the `.sur` io-plugin compatibility of Hyperspy 1.6, now Attolight files can be loaded using the `.sur` filetype.

Some protocols are being cleaned up to have external parameters, rather than hard-coded Attolight-specific parameters.

This PR aims at making Lumispy more general for any CL type of data, regardless of the acquisition system.

TODO:

- [x] `CLSEMSpectrum` class, fix the `correct_grating_shift` function
- [x] Write tests for each function


jlaehne:
Looks good to me.

The Attolight demo notebook fails reading the `.bin` files, but if I understand correctly, they would now have to be replaced by files in the `.sur` format?

Otherwise, it could be good if someone working with the Attolight system would have a look, can you ask e.g. Gunnar?

jordiferrero:
That's correct @jlaehne 
In order to read Attolight files now one must use `.sur` files instead.
They contain the same information as the `.bin` with additional metadata.

PS: I work with Gunnar :-) so he's on the loop.

PR #40, opened 2020-10-04
Functions adapted to hyperspy-based methods

jordiferrero:
Some functions currently in LumiSpy have created their own methods while there are similar functionalities in Hyperspy which can do the same.
In this PR, a better integration of Hyperspy-based methods is intended.

TODO:
- [ ] Cosmic ray removal function
- [ ] Background subtraction methods
- [ ] Create `utils/utils_xxx.py` files
- [ ] Create tests for each function


jordiferrero:
Closed as #44 was created.

PR #44, opened 2020-11-11
Functions adapted to hyperspy-based methods

jordiferrero:
Some functions currently in LumiSpy have created their own methods while there are similar functionalities in Hyperspy which can do the same.
In this PR, a better integration of Hyperspy-based methods is intended.
(Former #40)

TODO:
- [x] Cosmic ray removal function
- [x] Background subtraction methods
- [x] Create tests for each function


jordiferrero:
I am struggling with an error I get when running the Tests:
```
spikes_removal = SpikesRemoval(signal=signal, navigation_mask=navigation_mask, signal_mask=signal_mask,
                                       threshold=threshold, default_spike_width=default_spike_width,
>                                      add_noise=add_noise,)
E       TypeError: __init__() got an unexpected keyword argument 'threshold'
```

However, the `SpikesRemoval` class, inherited from `hyperspy.signal_tools` [here](https://github.com/hyperspy/hyperspy/blob/5d79acecada98a88899030bfd8a47d8e5bb58d5b/hyperspy/signal_tools.py#L1534), does indeed have a threshold argument. I don't know why it is raising this issue.

Can you @ericpre @jlaehne have a quick look if you see any obvious mistake? Thank you.

After resolving this issue, this PR is ready for review.

ericpre:
> I am struggling with an error I get when running the Tests in [this function](https://github.com/LumiSpy/lumispy/blob/d2607c60ef00647ef5725460893a7974d6d28faa/lumispy/signals/cl_spectrum.py#L109):
> 
> ```
> spikes_removal = SpikesRemoval(signal=signal, navigation_mask=navigation_mask, signal_mask=signal_mask,
>                                        threshold=threshold, default_spike_width=default_spike_width,
> >                                      add_noise=add_noise,)
> E       TypeError: __init__() got an unexpected keyword argument 'threshold'
> ```
> 
> However, the `SpikesRemoval` class, inherited from `hyperspy.signal_tools` [here](https://github.com/hyperspy/hyperspy/blob/5d79acecada98a88899030bfd8a47d8e5bb58d5b/hyperspy/signal_tools.py#L1534), does indeed have a threshold argument. I don't know why it is raising this issue.
> 
> When I run the tests internally in my computer (py3.7 in PyCharm) all tests pass...
> 
> Can you @ericpre @jlaehne have a quick look if you see any obvious mistake? Thank you.
> 
> After resolving this issue, this PR is ready for review.

This is not released and this is only on the `RELEASE_next_minor` branch.

lumispy/signals/cl_spectrum.py
```diff
@@ -33,49 +35,91 @@ class CLSpectrum(LumiSpectrum):
...
+        if inplace:
+            signal = self
+        else:
+            signal = self.deepcopy()
+
+        spikes_removal = SpikesRemoval(signal=signal, navigation_mask=navigation_mask, signal_mask=signal_mask,
+                                       threshold=threshold, default_spike_width=default_spike_width,
+                                       add_noise=add_noise,)
+
+        setattr(spikes_removal, 'noise_type', noise_type)
```
ericpre, line 113: Would it make sense to [set the noise in the metadata](https://hyperspy.readthedocs.io/en/latest/user_guide/signal.html#setting-the-noise-properties) instead?

And then rely on:
https://github.com/hyperspy/hyperspy/blob/1ff7468e41faa1094c189e4d7c182216efd41390/hyperspy/signal_tools.py#L1571-L1581

lumispy/signals/cl_spectrum.py
```diff
@@ -33,49 +35,91 @@ class CLSpectrum(LumiSpectrum):
...
+
+        if show_diagnosis_histogram:
+            self.spikes_diagnosis(navigation_mask=navigation_mask, signal_mask=signal_mask,
+                                  **kwargs)
+        if inplace:
+            signal = self
+        else:
+            signal = self.deepcopy()
+
+        spikes_removal = SpikesRemoval(signal=signal, navigation_mask=navigation_mask, signal_mask=signal_mask,
```
ericpre, line 109: Why not calling the `spikes_removal_tool` instead after setting the mask? If someone, wants to use the GUI, it would come for free? You could possibly reuse the docstring of this method too and adding the specific parameters of this method.

ericpre:
If you check the `py3.8/coverage` which uses the `non_uniform` branch and is ahead of (include) `RELEASE_next_minor`, you will notice that there is no error, but the test fails.

jordiferrero:
Oh. I assumed it was released already :-) Since I am always in the `non_uniform` branch....

So how should I deal with this? 
My guess it to add a check whether `SpikesRemovalTool` is released. If yes, then the function can be used. If not, then it is not available yet for the user until the hyperspy new version is released (I could leave my old simpler function for such cases as a temporal thing, but I would rather not). 
Does this make sense @ericpre? Thanks for the help!

lumispy/signals/cl_spectrum.py
jordiferrero: Yes. Indeed

lumispy/signals/cl_spectrum.py
jordiferrero: Mainly, because there is an issue in the `spikes_removal_tool`. I agree though, so I have fixed the function in PR:https://github.com/hyperspy/hyperspy/pull/2626.
I will fix the docstrings easily.

jlaehne:
`spikes_removal_tool` is an old function, which is released: https://hyperspy.org/hyperspy-doc/current/api/hyperspy._signals.signal1d.html?highlight=spikes_removal_tool#hyperspy._signals.signal1d.Signal1D.spikes_removal_tool

However, the `threshold` parameter was only introduced in https://github.com/hyperspy/hyperspy/pull/1412, which was merged last September. So you could maybe check for the availability of that parameter. As the use of lumispy is still limited without the non_uniform_axes, I would not bother providing alternative implementations.

jordiferrero:
All comments have been adressed and accepted.
This PR has been closed.
Thank you all for your help.

PR #45, opened 2020-11-17
Add function to join several Signal1D objects along signal axes

jlaehne:
Addes `join_spectra` function to `utils/axes`, which allows to join spectra with partially overlapping signal axes.

### Progress of the PR
- [X] Change implemented,
- [x] update docstring (if appropriate),
- [x] add tests,
- [x] ready for review.


jordiferrero:
Thank you Jonas.

Tests pass well. 
Could you perhaps add in the docstrings:
1. If it works for non-linear-axes
2. An example use case

For the rest, as soon as you tick all the progress checks I will merge.

jlaehne:
Good points, thanks for reminding me.

Should default to uniform axis if non-uniform-axes is not available, but includes handling for these already.

PR #43, opened 2020-11-02
implemented wavenumber conversion

jlaehne:
Extend the axis conversion utility to also handle wavenumbers (units of inverse cm) similar to the energy conversion from #37.

Includes two helper functions `nm2invcm` and `invcm2nm` for back and forth calculation.

### Progress of the PR
- [X] Change implemented in analogy to #37,
- [X] Raise error when non_uniform_axes functionality is not available from hyperspy,
- [x] Add function to calculate axis as relative wavenumber for given laser wavelength
- [X] update docstring (if appropriate),
- [x] add tests,
- [x] ready for review.

PR #46, opened 2020-11-30
JoinSpectra: Add error if scaling factor negative ...

jlaehne:
In a spectrum with bad background subtraction, I found that negative factors should be caught. Some more enhancements

Also added some corrections to the tests, as the tests for exceptions were only testing for the first type of excemption.

### Progress of PR:
- [X] Change implemented
- [X] Resolve issue with indices around center
- [X] gradient in averaging on overlap region
- [x] make scaling optional
- [x] Tests written and optimized using `mark.parametrize` (does that improve coverage?)
- [x] Ready for review


jlaehne:
Optimized the tests to make sure all parameter combinations are testes, but still coverage decreases. So as for #43, no idea why coverage decreases so drastically.

jlaehne:
Hm, coverage failed, but if you look at the details it worked well in fact.

Missing lines are mostly fixed in #43.

Only thing is handling of FunctionalDataAxis, which is an upstream bug in HyperSpy and I will fix it there.

jordiferrero:
That's good.
I merge the PR

PR #48, opened 2020-12-04
Test against hyperspy release or dev

ericpre:
Run test suite with the `non-uniform-axis` branch of hyperspy only for the coverage build. Other build use the release version of hyperspy.


ericpre:
@jordiferrero, @jlaehne

jordiferrero:
Thanks for that.
Why does the coveralls fail in this PR though? It should still pass, right?

ericpre:
If you compare https://coveralls.io/jobs/71749695 with https://coveralls.io/github/LumiSpy/lumispy, the coverage is the same.

Issue #17, opened 2020-04-01
signals vs. _signals


jlaehne:
@dnjohnstone Do you know why in Hyperspy the signal classes are in `_signals`, while we have them in `signals`. Would it make sense to copy the structure here in that aspect?

ericpre:
Since things staring with `_` are considered as private, I believe the idea behind putting  signals in `_signals` is to make these modules private and these are added to the `api` module to make them public: this is what we do when running `import hyperspy.api as hs`. Having an API module allows to explicitly define what is part of the public API or not.
This was behind "my time" as hyperspy contributor so I don't know the details, but @francisco-dlp will of course!
In principle, you can put signals anywhere you want but having a API consistent with other libraries doesn't sounds like a bad idea! As it is at the moment, it looks sensible to me.

francisco-dlp:
@ericpre, you guessed right. The idea was indeed not to expose the real location of the signals code in the pubic API so that changes in there don't break people's scripts and external packages.

jlaehne:
@jordiferrero, I think from the comments above, we can leave it as is.

The idea of LumiSpy is to be an extension of HyperSpy. As such, signals will likely not be called from LumiSpy directly, but as they are registered with HyperSpy the idea is to call them through the HyperSpy API.

It makes sense that we call the folder signals for clarity, but it does not need to be a private folder. I could though do the conversion to `_signals` if that would be of advantage, any opinion @ericpre, @jordiferrero ?

ericpre:
It may be more simple to keep `signals` and tidy up the namespace: currently, everything is imported in lumispy and with time, it can get cluttered.
Also the `load` function and the `roi` module is imported from hyperspy and added to the lumispy namespace, this is not a standard practise in python and can be confusing for users to have the same objects which can be imported from different libraries. For example, pxyem has standardised/simplified its namespace in https://github.com/pyxem/pyxem/pull/704/files and I think it would be good to do something similarly for lumispy.

jlaehne:
Good point, we mirrored the pyxem namespace, so now that it is cleaned up, we should do the same ;-)

I'll give it a try.

jlaehne:
Done with #49.

PR #49, opened 2021-01-27
Clean up namespace

jlaehne:
Based on the comment from at https://github.com/LumiSpy/lumispy/issues/17#issuecomment-767030637 I cleaned up the namespace in analogy to https://github.com/pyxem/pyxem/pull/704.

Progress:
- [X] Changes implemented
- [X] Adapted tests
- [x] Check again
- [x] Ready for review

Example:
Signals would be initiated as in HyperSpy and pyxem: 


jlaehne:
I'm thinking of joining all the three CL related files in `signals` into `cl_spectrum.py`.  I do not think we need to have so many different files, what do you think @jordiferrero @dnjohnstone . I've seen that in pyxem quite a number of different (though related) signal types are in one file.

jordiferrero (approved):

jordiferrero (approved):

jordiferrero:
> I'm thinking of joining all the three CL related files in `signals` into `cl_spectrum.py`. I do not think we need to have so many different files, what do you think @jordiferrero @dnjohnstone . I've seen that in pyxem quite a number of different (though related) signal types are in one file.

I am not sure about this...
When we discussed the initial signals structure, we realised that we needed some CL signals as some functions are only relevant for SEM and others are more general. I personally only work with SEM_CL data and like to have a special file with the functions for it (it's easier to navigate through classes).

I assume you are proposing to have the 3 classes in one file, right?
Happy to discuss further, but I would not merge them.

jlaehne:
Yep, that is exactly what I was thinking. 3 classes in one file. So we would have files for EL, PL, CL, but subclasses would be in the same.

I don't think there will be so many functions that are specific only to SEM or TEM. So far I mostly still have my signals as CLSpectrum.

We could also merge them now and if it ever gets too crowded still separate them.

Just some ideas, I am not passionate about either solution.

jordiferrero:
@jlaehne I agree to go ahead and merge the 3 classes in one file.
If you update this PR with those changes we can aprove it.

jlaehne:
done

(I tried to introduce a test for a warning from #44, but that would require the github-actions to include a hyperspy-gui package, so it is commented out for the moment. We can add that in a later PR to improve coverage.)

PR #55, opened 2021-03-05
add plot_linescan utility

jlaehne:
### Description of the change
Add a function to plot spectral linescans as colormap based on my stock function. Using pcolormesh, it can handle non-uniform axes.

### Progress of the PR
- [ ] add tests,
- [ ] ready for review.

```python
import lumispy as lum
import numpy as np
s = lum.signals.LumiSpectrum(np.random.rand(10,100)*10)
lum.plot_linescan(s)
```

### DIFF
--- a/lumispy/__init__.py
+++ b/lumispy/__init__.py
@@ -22,6 +22,7 @@ import logging
 _logger = logging.getLogger(__name__)
 
 from lumispy.utils.axes import nm2eV, eV2nm, nm2invcm, invcm2nm, join_spectra
+from lumispy.utils.plot import plot_linescan
 
 from lumispy import signals
 from lumispy import components
--- a/lumispy/utils/plot.py
+++ b/lumispy/utils/plot.py
@@ -1,0 +1,166 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019-2021 The LumiSpy developers
+#
+# This file is part of LumiSpy.
+#
+# LumiSpy is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# LumiSpy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with LumiSpy.  If not, see <http://www.gnu.org/licenses/>.
+
+import numpy as np
+import matplotlib.pyplot as plt
+from matplotlib import ticker
+from scipy.interpolate import interp1d
+
+
+def plot_linescan(s, logscale=True, colorbar=True, invert=False, cmap='jet', \
+            filename=None, vmin=1, vmax=None, interpolate=None, contour=False, \
+            clevels=15, figsize=None, vline=None, vlinecolor='black', \
+            hline=None, hlinecolor='black', linewidth=1, fontsize=18, \
+            fontfamily='DejaVu Sans'):
+    """Plot spectral linescan as colormap on screen and optionally write to
+    file.
+    
+    Parameters
+    ----------
+    s : LumiSpectrum object
+        Works only for exactly one navigation dimension.
+    logscale : boolean, optional
+        If `True` (default), plot on a logarithmic intensity scale.
+    colorbar: boolean, optional
+        If `True` (default), add colorbar to the right.
+    invert: boolean, optional
+        Invert navigation direction. Default `False`.
+    cmap : string, optional
+        Color map, e.g. 'jet' (default), 'inferno', ...
+    filename : string, optional
+        If not `None, write to an image file with this filename. The format is
+        determined by the extension, e.g. 'outfile.png'.
+    vmin : float, optional
+        Minimum `z` (intensity) value. Default 1.
+    vmax : float, optional
+        If `None` (default), automatically determine the maximum value of the
+        `z` (intensity) scale from dataset. Otherwise take given value.
+    interpolate : int
+        If not `None` (default), interpolate to finer mesh with given number of
+        interpolation points both for x and y direction.
+    contour : boolean, optional
+        Plot additional contour lines (default `False`)
+    clevels : int
+        Number of contour levels (default 15)
+    figsize : tuple of floats
+        If `None` (default) use standard figure size, otherwise use size
+        defined by the given tuple.
+    vline : float or array of floats
+        Position(s) along signal axis at which to plot a vertical line.
+        Default `None` corresponds to no lines.
+    vlinecolor : string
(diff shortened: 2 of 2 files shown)


ericpre:
An alternative would be to add support for non uniform axis to the `hs.plot.plot_spectra`, wouldn't it be more simple?

jordiferrero:
> An alternative would be to add support for non uniform axis to the `hs.plot.plot_spectra`, wouldn't it be more simple?

I agree. `hs.plot.plot_spectra(s, style='heatmap')` can do the same job.

Unless we wanted to add other functionalities to such plots (e.g. markers which follow peaks), I would take @ericpre approach. 
I am not sure though of the complexity of `hs.plot.plot_spectra`. I can try to have a look and see where in the code it fails with non-linear axes...

jlaehne:
Hyperspy `plot_spectra` and other plot functions would need to use `pcolormesh`  to correctly render nonlinear axes. It's on my todo-list to have a look at those. My linescan utility has vertical/horizontal lines, contour lines and interpolation as additional optional features that I don't think are available in hyperspy. But as I had written this function over the years explicitly with non-linear axes in mind, I had never realized that plot_spectra also does linescans. I'll have a closer look at that function again to propose a way to go forward.

jlaehne:
Example using log-scale, with contours and marker lines:
image

Same data with different figure size, inverted navigation axis, linear scale, different cmap, energy axis:
image

ericpre:
> `plot_spectra` from HyperSpy is also missing a `figsize` argument and uses a non-optimal aspect ratio, which would have to be tweaked to be comparable. At the moment, the `cmap` argument is also not working.
> 
> Adding the parameters the `plot_linescan` function provides to `plot_spectra` would overload it in my eyes (even though it might be further improved to use `pcolormesh` at least).

These should be easy to fixed, the `hs.plot.plot_spectra` function is fairly simple.
Also it would be good to have the contour functionality as a separate functions acting on matplotlib axes, so that it can be re-use by other functions to plot images, for example.

jlaehne:
OK, if you prefer to extend `plot_spectra` with these extra functionalities, I can integrate the functionalities from here in that function. Adding lines can be an interesting option for spectra as well.

jordiferrero:
> OK, if you prefer to extend `plot_spectra` with these extra functionalities, I can integrate the functionalities from here in that function. Adding lines can be an interesting option for spectra as well.

@ericpre Did you mean to add these functions in the hyperspy package or here?

I agree with you that the hyperspy `plot_spectra` is a bit complicated and not particularly useful for fine control... I found myself many times trying to change parameters that could not be accessible with it... and then creating my plotting functions myself and extracting the axes and data manually...


PS:
For the figsize issue, isn't it as simple as plotting with the `fig` argument (I am not sure but similar things have worked for me before):
```
f, axs = plt.subplots(figsize=(5,4))
hs.plot.plot_spectra(s, fig=f)
```

jlaehne:
I can have a look to improve `plot_spectra` to see if I manage to incorporate these things there. Some more versatility would probably be helpful to a lot of people. I also have my own `plot_spectra` function for some functionalities I was missing in the hyperspy one.

I have kept them separate not to overload a single function, but it is true that a stack/list of single spectra or a linescan do have some things in common and it might not be bad to improve this 'super' function.

Apart from that, we could still decide to have this as a convenience function in LumiSpy. This function is what I have been developing over the years and in itself is quite mature for linescans. Only thing that would need to be added at the moment are the tests.

ericpre:
You will be surprise how much functionalities you can pack in a function! ;) `hs.plot.plot_images` is quite bad but `hs.plot.plot_spectra` is quite simple and from the discussions here, there are quite a bit of room for improvement.

For example:
```python
import hyperspy.api as hs

s = hs.datasets.artificial_data.get_core_loss_eels_line_scan_signal()
for style in ['overlap', 'cascade', 'mosaic', 'heatmap']:
    hs.plot.plot_spectra(s, style=style)
```
and [this example](https://hyperspy.readthedocs.io/en/latest/user_guide/visualisation.html#plotting-profiles-interactively) which I particularly like for its simplicity!

Generally, I find it confusing when several functions/method do very similar kind of things... because it is not clear which functions should be used, the difference, etc... For example, in the last few days, I had to do some interpolation and I find it confusing that there are many functions in [scipy.interpolate](https://docs.scipy.org/doc/scipy/reference/interpolate.html). Obviously, most of them are different but for some of them, it is difficult to know what are they differences by reading the documentation!

jordiferrero:
Any updates on this?
Judging from #139 and this PR, it seems that plotting utilities specific for luminescence data always get stuck in PR revision because, in theory, "it can be done with the default hyperspy functions". But doesn't this argument also make the `hs.plot` functions redundant because matplotlib can do the exact same functionality?

What I am trying to say here @ericpre and @jlaehne is that lumispy could have luminescence-specific plotting utilities (in `utils/plot` tailored to some key functionalities that are maybe less relevant for EDX or other types of spectroscopic. Maybe some plotting utils make some of the lumispy-demos easier to read/interact with... I don't think that would be terrible.
On the other hand, I do see the downsides of having very similar functions in different packages.

jlaehne:
> Any updates on this?

I never got around looking at the HyperSpy functions in detail and how to implement the functionality there. Indeed, we should discuss whether we might not have a few specific extra functions here. So far I simply still use my offline version of that function, from which this PR was forked.

It does make sense to give HyperSpy functions some extra functionality, but when they get more complicated also the barrier to use them for specific types of plots gets more difficult. Thinking about it, I do still see an extra benefit of a dedicated linescan plotting function with a very dedicated set of options.

If we should agree on that, I would go over the PR again as some of the arguments could probably be named more close to `matplotlib` standards, when I originally wrote the code I just used what came to my mind.

jordiferrero:
> > Any updates on this?
> 
> I never got around looking at the HyperSpy functions in detail and how to implement the functionality there. Indeed, we should discuss whether we might not have a few specific extra functions here. So far I simply still use my offline version of that function, from which this PR was forked.
> 
> It does make sense to give HyperSpy functions some extra functionality, but when they get more complicated also the barrier to use them for specific types of plots gets more difficult. Thinking about it, I do still see an extra benefit of a dedicated linescan plotting function with a very dedicated set of options.
> 
> If we should agree on that, I would go over the PR again as some of the arguments could probably be named more close to `matplotlib` standards, when I originally wrote the code I just used what came to my mind.

I would go ahead. But let's wait for @ericpre opinion on having some lumispy-specific plotting utils

ericpre:
As far as I understand, the functionalities discussed here are useful for most spectroscopy technique and I don't see what is luminescence specific!

jlaehne:
> As far as I understand, the functionalities discussed here are useful for most spectroscopy technique and I don't see what is luminescence specific!

Well, for e.g. an EDX linescan, you would normally not plot the full spectra against position (might look fun though). For other signal types used in HyperSpy and not included in LumiSpy it may indeed be useful as well. We simply use linescans quite a lot and this is why such a tool can be handy - in contrast to `plot_spectra`, where this way of displaying would just be an extra among different functionalities.

Issue #52, opened 2021-02-07
loadtxt, savetxt


jlaehne:
I want to implement a simple set of wrappers that use np.loadtxt and np.savetxt to load and save txt, dat, csv format files. Mostly metadata agnostic of course.  At least for single spectra as well as linescans. For more than one navigation dimension it would become more complicated, but at least 2d navigation with 0d signal axis (e.g. single parameters from fit results) would also be feasible. It would be helpful to import/transfer data from systems where the native file system is so far not implemented in hyperspy (e.g. Jobim Yvon spectrometer software or systems with self-programmed data collection), but a text format export is possible. On the writer side, I often want to pass on some data to colleagues who do not use hyperspy or even python (the good old Origin crowd). 

@ericpre, @dnjohnstone, @jordiferrero would it make more sense to implement that in the io-part of hyperspy for general usage? Otherwise I would implement a specific set of functions just for lumispy?

ericpre:
> I want to implement a simple set of wrappers that use np.loadtxt and np.savetxt to load and save txt, dat, csv format files. Mostly metadata agnostic of course.

The issue with having a generic io plugin for txt/csv files is that there is no specification and therefore the plugin can't do much:
- loading data is trivial
- nothing to be done with restructuring the data
- not possible to set axes
- not possible to add metadata.

> It would be helpful to import/transfer data from systems where the native file system is so far not implemented in hyperspy (e.g. Jobim Yvon spectrometer software or systems with self-programmed data collection), but a text format export is possible.

If there are specifications or it follows a specific know pattern, then this is different from the case above and, it makes sense to have a plugin - see for example [`msa`](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/io_plugins/msa.py).

ericpre:
Having examples would be good, see
https://github.com/hyperspy/hyperspy/pull/2306/files#diff-82b6f06aa6e3647b596d6dc24789557125d85d87ce203ad9a660ffc5c5ec78f4

jlaehne:
I see your point with the 'specs'. At least for saving I have my wrapper routines and think could make sense

jordiferrero:
I agree that we need some sort of function to integrate better with the Origin crowd.
Gunnar (here in Cambridge) has asked me for this functionality many times, as collaborators may not use Python.

[...]

PR #56, opened 2021-03-08
Savetxt

jlaehne:
### Description of the change
Proposal for a `savetxt` function based on the discussion in #52. Functionality to write simple text files for data transfer for signals that have at most two axes (signal or navigation axes). Use cases would be especially spectral linescans or maps of fit parameters to pass them to colleagues who prefer `txt` files over hdf5 files.

### Progress of the PR
- [x] Make saving axes optional and add transpose option,
- [x] update docstring (if appropriate),
- [x] add tests,
- [x] ready for review.

```python
import lumispy as lum
import numpy as np

# Spectrum:
s = lum.signals.LumiSpectrum(np.arange(5))
s.savetxt('spectrum.txt')
# 0.00000	0.00000
# 1.00000	1.00000
# 2.00000	2.00000
# 3.00000	3.00000
# 4.00000	4.00000

# Linescan:
l = lum.signals.LumiSpectrum(np.arange(25).reshape((5,5)))
l.savetxt('linescan.txt')
# 0.00000	0.00000	1.00000	2.00000	3.00000	4.00000
# 0.00000	0.00000	5.00000	10.00000	15.00000	20.00000
# 1.00000	1.00000	6.00000	11.00000	16.00000	21.00000
# 2.00000	2.00000	7.00000	12.00000	17.00000	22.00000
# 3.00000	3.00000	8.00000	13.00000	18.00000	23.00000
# 4.00000	4.00000	9.00000	14.00000	19.00000	24.00000
```


jordiferrero:
Thanks for starting this up. Looks good so far.
Can you please add how the output of the function would look in the case example at the description?

jlaehne:
I improved the example above and included that in the docstring using the formalism discussed in #58 (though without putting the docstrings in an extra folder).

lumispy/signals/luminescence_spectrum.py
```diff
@@ -344,6 +345,45 @@ def remove_background_from_file(self, background=None, inplace=False, **kwargs):
...
+    """
+
+    def savetxt(self, filename, fmt='%.5f', delimiter='\t', axes=True,
+                transpose=False, **kwargs):
+        """Writes luminescence spectrum object to simple text file.
+        %s
+        %s
+        %s
+        """
+        savetxt(self, filename, fmt, delimiter, axes, transpose, **kwargs)
```
jordiferrero, line 382: I think these function should be called slightly different. I find it confusing that the utils and the LumiSpectrum function have the same name...

lumispy/utils/io.py
```diff
@@ -0,0 +1,131 @@
...
+    %s
+    """
+    nav_axes = S.axes_manager.navigation_axes
+    sig_axes = S.axes_manager.signal_axes
+    dim = len(nav_axes) + len(sig_axes)
+    # Write single spectrum or data with only navigation axis
+    if dim == 1:
+        if axes:
+            if len(sig_axes) == 1:
+                X = sig_axes[0].axis
```
jordiferrero, line 97: The general convention is lower case for variables and Capitalised case for classes.
Consider changing `X` for `x` and `S` for s...

lumispy/utils/io.py
```diff
@@ -0,0 +1,131 @@
...
+            else:
+                X = nav_axes[0].axis
+            output = np.array([X,S.data]).T
+        else:
+            output = S.data
+        np.savetxt(filename, output, fmt=fmt, delimiter=delimiter, **kwargs)
+    # Write linescan or matrix
+    elif dim == 2:
```
jordiferrero, line 105: I am not sure (and I am happy for you to keep this like that, but what about creating 2 sub-functions for each case?
So:
```
if dim ==  1:
    savetxt_1d(*)
elif dim == 2:
     savetxt_2d(*)
```
This would allow us to add other dimensions in the future in a clean and easy way?

lumispy/signals/luminescence_spectrum.py
```diff
@@ -344,6 +345,45 @@ def remove_background_from_file(self, background=None, inplace=False, **kwargs):
...
+    >>> l.savetxt('linescan.txt')
+    # 0.00000	0.00000	1.00000	2.00000	3.00000	4.00000
+    # 0.00000	0.00000	5.00000	10.00000	15.00000	20.00000
+    # 1.00000	1.00000	6.00000	11.00000	16.00000	21.00000
+    # 2.00000	2.00000	7.00000	12.00000	17.00000	22.00000
+    # 3.00000	3.00000	8.00000	13.00000	18.00000	23.00000
+    # 4.00000	4.00000	9.00000	14.00000	19.00000	24.00000
+    """
+
+    def savetxt(self, filename, fmt='%.5f', delimiter='\t', axes=True,
```
jordiferrero, line 375: That's a personal thought but worth mentioning...
I like to keep my data in pandas dataframes many times as I love using seaborn for plotting trends.

Could you add a parameter (if you think it's not too confusing) called `return_pandas_df` which returns the Pandas dataframe object  instead of saving a txt file (the same as you would save as txt).
That would save me to have to run two lines of code like that:

```
cl.savetxt('file.txt')
df = pd.read_csv('file.txt')
```

jordiferrero (changes requested):

lumispy/signals/luminescence_spectrum.py
jlaehne: Well, they do exactly the same - except that the utils-functions also accepts other hyperspy signal classes than LumiSpectrum. The latter is more for convenience. So I woul prefer keeping the same name.

lumispy/utils/io.py
jlaehne: It is not so much code that I consider it worth splitting, and even a third dimension could be easily added - anything beyond will get hard to handle in a text file.

lumispy/signals/luminescence_spectrum.py
jlaehne: sure, I can do so - then we can take pandas conventions as a guideline for the third dimension.

jlaehne: I would maybe call it `return_array` and then you would have to run the line 
```
df = pd.DataFrame(cl.savetxt(return_array=True))
```

This avoids introducing pandas as dependency, but pandas could be mentioned as use case in the documentation.

Though it is a bit weird to run a `savetxt` function for that. Alternatively, I could put that part into a `to_array` function (open for better name proposals) with the same parameters that is called by `savetxt` but can also be called separately for the pandas use case. What do you think?

jordiferrero: I agree that we may be mixing functions... Let's forget about pandas for now I guess.
Maybe if you managed to split the save_txt function as a `to_array` + `savetxt` functions, then we could just call the first half of the workflow?
Does that make sense?

jlaehne: That is what I thought, see update

jordiferrero (approved):

PR #58, opened 2021-03-09
Grating solver

jordiferrero:
### Description of the change
Added a solver for the grating/spectrometer equation. 
- A `utils.axis` function that converts an axis in pixel/undefined units to wavelength in nanometers
- A `px_to_nm_grating_solver` in the `LumiSpectrum` class that converts the object's signal axis to wavelength in nanometers.

### Progress of the PR
- [x] Change implemented
- [x] Add tests,
- [x] ready for review.

```python
import lumispy as lum
import numpy as np
s = LumiSpectrum(np.ones(20),))
s.px_to_nm_grating_solver(*exp_paramers , inplace=True)
s.axes_manager.signal_axes[0].units

>> 'nm'
```

### DIFF
--- a/lumispy/signals/luminescence_spectrum.py
+++ b/lumispy/signals/luminescence_spectrum.py
@@ -344,7 +344,7 @@ class LumiSpectrum(Signal1D, CommonLumi):
                            grating_central_wavelength_nm, grating_density_gr_mm, inplace=False,):
         """
         Converts signal axis of 1D signal (in pixels) to wavelength, solving the grating
-        equation. See
+        equation. See ``lumispy.axes.solve_grating_equation` for more details.
         Input parameters
         ----------------
         :param gamma_deg:
@@ -352,7 +352,7 @@ class LumiSpectrum(Signal1D, CommonLumi):
             (found experimentally from calibration). In degree.
         :param deviation_angle_deg:
             Also known as included angle. It is defined as the difference between
-            angle of diffraction ($\beta$) and angle of incidence (&\alpha$).
+            angle of diffraction ($\beta$) and angle of incidence ($\alpha$).
             Given by manufacturer specsheet. In degree.
         :param focal_length_mm:
             Given by manufacturer specsheet. In mm.
@@ -378,18 +378,19 @@ class LumiSpectrum(Signal1D, CommonLumi):
                                          gamma_deg, deviation_angle_deg, focal_length_mm, ccd_width_mm,
                                          grating_central_wavelength_nm, grating_density_gr_mm)
 
-        # in place conversion
         if inplace:
-            self.axes_manager.remove(-1)
-            self.axes_manager._axes.append(nm_axis)
-            return
-
-        # create and return new signal
+            s = self
         else:
             s = self.deepcopy()
-            s.axes_manager.remove(-1)
-            s.axes_manager._axes.append(nm_axis)
+
+        s.axes_manager.remove(-1)
+        s.axes_manager._axes.append(nm_axis)
+
+        if inplace:
+            return
+        else:
             return s
+
 
 
 class LazyLumiSpectrum(LazySignal, LumiSpectrum):
--- a/lumispy/signals/luminescence_spectrum.py
+++ b/lumispy/signals/luminescence_spectrum.py
@@ -18,7 +18,7 @@
 
 """Signal class for Luminescence spectral data (1D).
 """
-
+import warnings
 from inspect import getfullargspec
 from warnings import warn
 import numpy as np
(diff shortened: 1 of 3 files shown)


jordiferrero:
I do have a few comments/questions:
1. What is the best way to "inherit" docstrings? Let me explain. In this PR I have 2 functions, one stored in `utils.axis` and the other one is the method that calls the `util` function though the `LumiSpectrum` class. Both functions have the same parameters (`*exp_parameters`). They are described manually in both functions. Is there an easy way to inherit the docstrings for the `utils` function onto the `LumiSpectrum` function?

2. I am assuming that the refraction index in air is 1, but that is not "exact". Do you think I should not assume so? 

Thank you!

jlaehne:
>     1. What is the best way to "inherit" docstrings? Let me explain. In this PR I have 2 functions, one stored in `utils.axis` and the other one is the method that calls the `util` function though the `LumiSpectrum` class. Both functions have the same parameters (`*exp_parameters`). They are described manually in both functions. Is there an easy way to inherit the docstrings for the `utils` function onto the `LumiSpectrum` function?
Good point, would apply to #56 as well.

>     2. I am assuming that the refraction index in air is 1, but that is not "exact". Do you think I should not assume so?
Well it should be easy to use the function `_n_air(wl)` to correct for that (for the range 185 nm to 1700 nm). Why not use it even if the improvement is probably not so great. The error introduced by the abberation of the spectrometer will probably be larger, but that is another story.

lumispy/signals/luminescence_spectrum.py
```diff
@@ -340,6 +340,58 @@ def remove_background_from_file(self, background=None, inplace=False, **kwargs):
                 return self.map(lambda s, bkg: s - bkg, bkg=bkg_y, inplace=True)
 
 
+    def px_to_nm_grating_solver(self, gamma_deg, deviation_angle_deg, focal_length_mm, ccd_width_mm,
+                           grating_central_wavelength_nm, grating_density_gr_mm, inplace=False,):
```
jlaehne, line 344: parameter names are fairly long, on the other hand they are very explicit, which is nice

lumispy/signals/luminescence_spectrum.py
```diff
@@ -340,6 +340,58 @@ def remove_background_from_file(self, background=None, inplace=False, **kwargs):
...
+
+        """
+
+        nm_axis = solve_grating_equation(self.axes_manager.signal_axes[0],
+                                         gamma_deg, deviation_angle_deg, focal_length_mm, ccd_width_mm,
+                                         grating_central_wavelength_nm, grating_density_gr_mm)
+
+        # in place conversion
+        if inplace:
+            self.axes_manager.remove(-1)
```
jlaehne, line 383:
```
if inplace: s = self
else: s = self.deepcopy()

s.axes_manager.remove(-1)
s.axes_manager._axes.append(nm_axis)
return s
```

saves you a few duplicate lines

jordiferrero:
@jlaehne Most of these "aesthetic" changes I did not add them myself manually... 
I use PyCharm and it always corrects for them automatically, to keep things consistent, I suppose...
I personally find it confusing to not use espaces in some occasions while using them most of the time.

Thanks for the comments though!

lumispy/signals/luminescence_spectrum.py
```diff
@@ -378,20 +378,21 @@ def px_to_nm_grating_solver(self, gamma_deg, deviation_angle_deg, focal_length_m
...
-            s.axes_manager.remove(-1)
-            s.axes_manager._axes.append(nm_axis)
+
+        s.axes_manager.remove(-1)
+        s.axes_manager._axes.append(nm_axis)
+
+        if inplace:
+            return
+        else:
             return s
```
jlaehne, line 392: `return s` is sufficient in this case, as it is a pointer to self in case of `inplace`

ericpre:
> 1. What is the best way to "inherit" docstrings? Let me explain. In this PR I have 2 functions, one stored in `utils.axis` and the other one is the method that calls the `util` function though the `LumiSpectrum` class. Both functions have the same parameters (`*exp_parameters`). They are described manually in both functions. Is there an easy way to inherit the docstrings for the `utils` function onto the `LumiSpectrum` function?

Include string in the docstring, for the syntax see for example:
https://github.com/hyperspy/hyperspy/tree/RELEASE_next_minor/hyperspy/docstrings
and 
https://github.com/hyperspy/hyperspy/blob/165e407e777b503cdbc1b3d29dcf481ac15b7e4b/hyperspy/signal.py#L2866-L2901

jlaehne:
Thanks @ericpre, I already used that in an update to #56 @jordiferrero. Though I preferred to keep the docstring in the file where the function is defined rather than in an extra folder.

ericpre:
Yes, the string used by the docstring can live anywhere! I guess the reason why the docstring are in separate folder in hyperspy is that most of them are used across several files.

jordiferrero:
I have now included the changes to the dosctrings.
Ready for final review.

jlaehne (approved):
LGTM

PR #68, opened 2021-03-31
GitHub actions: Create release.yml

jlaehne:
### Description of the change
Setting up the github actions to automate releasing when a new release tag is pushed. Based on hyperspy's `release.yml`.

### Progress of the PR
- [X] Adapted file from `hyperspy_gui_ipywidgets`,
- [X] set pypi secrets,
- [X] test run in fork,
- [x] add MANIFEST.in
- [x] add changelog,
- [x] Add `releasing_guide.md` to have the recipe for a release summarized,
- [x] Add keywords to `release_info.py`,
- [x] ready for review.


.github/workflows/release.yml
```diff
@@ -0,0 +1,135 @@
...
+          python-version: 3.8
+
+      - name: install twine
+        run: |
+          python -m pip install twine
+      - name: Build manylinux Python wheels
+        uses: RalfG/python-wheels-manylinux-build@v0.3.3
+        with:
+          python-versions: 'cp36-cp36m cp37-cp37m cp38-cp38 cp39-cp39'
+          build-requirements: 'cython'
```
jlaehne, line 104: Do we need the cython build?

.github/workflows/release.yml
```diff
@@ -0,0 +1,135 @@
...
+          path: ./dist/*.whl
+
+      - name: Publish wheels to PyPI
+        if: github.repository_owner == 'lumispy'
+        env:
+          # Github secret set in the lumispy/lumispy repository
+          # Not available from fork or pull request
+          # Secrets are not passed to workflows that are triggered by a pull request from a fork
+          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+          TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
```
jlaehne, line 82: need to be added

.github/workflows/release.yml
```diff
@@ -0,0 +1,135 @@
...
+          python -m pep517.build --binary --out-dir dist/ .
+      - name: Display content dist folder
+        run: |
+          ls dist/
+      - name: Install and test distribution
+        env:
+          MPLBACKEND: agg
+        run: |
+          pip install --pre --find-links dist lumispy[tests]
+          pytest --pyargs lumispy
```
jlaehne, line 70: Do we need to add test against `non_uniform_axis` branch of hyperspy?

.github/workflows/release.yml
ericpre: No, this is not necessary here - for hyperspy, we need it for the bcf reader.

.github/workflows/release.yml
```diff
@@ -0,0 +1,135 @@
...
+      - name: Publish wheels to PyPI
+        if: github.repository_owner == 'lumispy'
+        env:
+          # Github secret set in the lumispy/lumispy repository
+          # Not available from fork or pull request
+          # Secrets are not passed to workflows that are triggered by a pull request from a fork
+          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+          TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+        run: |
+          twine upload dist/*-manylinux*.whl --verbose
```
ericpre, line 134: Depending on one of the comment above, this may need to be changed to:
```suggestion
          twine upload dist/*-py3-none-any.whl --verbose
```

ericpre (commented):
I have left a few comments.

.github/workflows/release.yml
```diff
@@ -0,0 +1,135 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: install twine
+        run: |
+          python -m pip install twine
+      - name: Build manylinux Python wheels
```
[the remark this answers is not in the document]
jlaehne: I simplified the whole workflow based on the `hyperspy_gui_ipywidgets` package

.github/workflows/release.yml
jlaehne: as above, similar as in ipywidgets *.whl is directly uploaded

ericpre:
Can you try to push a tag in your fork to check that it triggers the workflow and that everything goes well? Normally, everything should pass and the last step with pushing to pypi should be skipped.

jlaehne:
I did. It fails, but at the same point as for `hyperspy_gui_ipywidgets`: https://github.com/jlaehne/lumispy/runs/2249602259?check_suite_focus=true

So is that sufficient?

For the moment I dropped automatic writing of the version, because we have the version in `release_info.py` and as far as I see the solution from `hyperspy_gui_ipywidgets` overwrites the full file.

jlaehne:
Maybe we could stick with manually setting the version in release_info.py, but do a check in the github action that the version in the file (can be read as `$(python setup.py --version)`) is the same as in the tag?

.github/workflows/release.yml
jordiferrero: Sure. @jlaehne can you please create a user account in pypi and send me your user name?

jordiferrero: I have also added the TWINE_USERNAME and PASWORD as a secret for the repository.

ericpre:
Yes, what I have experimenting with updating the version is not great and it can be done manually as you suggest. There is an [setuptools-scm](https://pypi.org/project/setuptools-scm) which seems to be worth considering, but even if these things have changed fluid in the last few years, maybe it has now converged to something useful?!

In case of https://github.com/jlaehne/lumispy/runs/2249602259?check_suite_focus=true, this is failing at the very beginning when creating the github release and according to error message, the release already exists:
```
Error: Validation Failed: {"resource":"Release","code":"already_exists","field":"tag_name"}
```
Is it possible that it already exists from a previous run with the same tag?

.github/workflows/release.yml
jlaehne: done

jlaehne:
I tried with new tag, but the same: https://github.com/jlaehne/lumispy/actions/runs/711368311

ericpre:
I am not sure which branch which we branch I should be looking at, the one in `lumispy/lumispy` or in `jlaehne/lumispy`? This PR is from a branch from `lumispy/lumispy`.

ericpre:
I did the following:
- add a git tag (`v0.1.1.test`)
- push the tag to `ericpre/lumispy`

and it works fine: https://github.com/ericpre/lumispy/actions/runs/711407096

jlaehne:
> I am not sure which branch I should be looking at, the one in `lumispy/lumispy` or in `jlaehne/lumispy`? This PR is from a branch from `lumispy/lumispy`.

I first created the `lumispy/lumispy` PR and now used a copy in my repo only for playing around.

So far I had used `Draft a New Release` to create the tag. Doing it on the command line worked now: 
https://github.com/jlaehne/lumispy/actions/runs/711462046

jlaehne:
We still need to decide how to track the `CHANGES`. The solution in HyperSpy should be changed again, as far as I understood.

jlaehne:
@ericpre Do you have an idea how to simply verify that the version in `release_info.py` and the current tag are equal? Should be enough for versioning, but would be a sanity check.

ericpre:
> We still need to decide how to track the `CHANGES`. The solution in HyperSpy should be changed again, as far as I understood.

In hyperspy, this is annoying because there is more traffic and a number of PRs take more time to get merged, which is the source of conflict of the change log...
To have it automated, [towncrier](https://github.com/twisted/towncrier) seems a good option, it used by many repository on github and for example [pipenv](https://github.com/pypa/pipenv) uses it.

> @ericpre Do you have an idea how to simply verify that the version in `release_info.py` and the current tag are equal? Should be enough for versioning, but would be a sanity check.

As mentioned above, I would check if setuptools-scm can be useful for that.

jordiferrero:
Let me know if you need any help with this.
I am a bit unsure how to help if I am honest. But happy to learn!

jlaehne:
In principle the action is running. I was trying to add a consistency check between the version in the tag and the version in `release_info.py` as a condition for the rest of the action to run, but I have not really progressed on that so far.  I have not found time to dig deeper into `setuptools-scm` as alternative, but I thought we are not that far with the project that we really need those extras yet. We could live with ensuring the right versioning manually for the moment, I guess.

Concerning the changelog, there is the solution proposed by @ericpre using `towncrier`. But also there are a large number of Github actions that would compile a changelog from the PRs that have been merged, which would probably be enough automation for us. But then, which one to choose. In HyperSpy, the changelog file is updated manually, but that currently leads to a lot of merge conflicts that have to be resolved manually. I guess for the moment we could even live with that, but then it does not hurt to do things right.

Just had a look at pyxem and they simply do all that manually. So how about keeping these as nice to have features for the future if one of us finds the time to dig deeper? Then we can focus on adding other functionalities for the moment.

jordiferrero:
I'm happy with not worrying too much on this.
Since it's only 3 of us working on this project we should be fine for now.
I've added it in the project "Nice features for the future".
If you all agree, we can merge and close this for now.

ericpre:
> In principle the action is running. I was trying to add a consistency check between the version in the tag and the version in `release_info.py` as a condition for the rest of the action to run, but I have not really progressed on that so far.

Yes, this is good idea to check that there is no mistake in the version... I would write a python script comparing the tag, which is available as environment variable and compare it with the `lumispy.__version__`. Maybe it could fit in one python line, which could be added to the workflow without adding a file to lumispy.

Regarding the changelog, it can indeed be done manually!

ericpre (approved):
This looks good to me. I ran a workflow in https://github.com/ericpre/lumispy/actions/runs/711407096 and it went through until the end. The only thing left to check is if the secrets are set correctly for pypi and there is no other way than actually doing it! If it fails, this is not a big deal, just need to delete the tag and the github release and make sure if there are correct and try it again.

jordiferrero:
What do you mean by "another way to store the secrets"? Isn't the Github way the correct way?

jlaehne:
> Yes, this is good idea to check that there is no mistake in the version... I would write a python script comparing the tag, which is available as environment variable and compare it with the `lumispy.__version__`. Maybe it could fit in one python line, which could be added to the workflow without adding a file to lumispy.

In principle it should work like that  and I will give it a try as soon as I find a quiet moment, maybe later this week. There is just a bit of other stuff that has piled up over the easter holidays.

> Regarding the changelog, it can indeed be done manually!
Yes, I will simply create the file in this PR.

Once everything is in, I will do the next minor realease to test the workflow.

ericpre:
> What do you mean by "another way to store the secrets"? Isn't the Github way the correct way?

Yes, setting the github secrets is good, but since this is not possible to check after there have been set and it is always possible to have a typo or a mistake in the the copy and paste, etc. For example, it happen to me recently!

jlaehne:
Release workflow runs through when version is correct: https://github.com/jlaehne/lumispy/actions/runs/783664680
but fails if there is a mismatch between `release_info.py` and the `tag`: https://github.com/jlaehne/lumispy/actions/runs/783667655

jlaehne:
@jordiferrero should be ready now, though I do not know why the tests of the `grating_shift` suddenly fail, because they were not touched - could it be a result of the new patch release of HyperSpy @ericpre?

ericpre:
Indeed, the failure are related to changes in the `map` method:
```python
ValueError: The size of the navigation_shape for the kwarg shift (<(100,)> must be consistent with the size of the mapped signal <(10, 10)>
```
This point out to an argument which doesn't have the right shape. It is fairly likely that ambiguous usage of `map` such as this one are not allowed anymore.

jlaehne:
OK, I created a separate issue #75 for the failing tests.

This one should be ready then.

jlaehne:
By the way, I just registered [lumispy.org](http://lumispy.org), which forwards to this repository for the moment, but can be later used as link for the documentation.

Issue #65, opened 2021-03-27
LumiSpy logo


jlaehne:
Any ideas for a LumiSpy logo?

One idea I had was using this map of emission energy around a dislocation in GaN, but any further ideas are welcome.
![image](https://user-images.githubusercontent.com/7076057/112737796-5d997280-8f5d-11eb-92a4-bb02e852cde6.png)

jordiferrero:
Your last comment triggered one of my hobbies, computer design.
I made some sketches of ideas I also had a while ago combining the *lambda* and the *spectral colours* (like it is not that important... but it is fun!)

Happy to combine my ideas with your GaN, let me know what you think :-)
<img width="3043" alt="Asset 15@2x" src="https://user-images.githubusercontent.com/9082687/112761294-edd9c500-8ffa-11eb-85c2-6a4e34f0db2c.png">

ericpre:
@jordiferrero, some very good ideas in your design! Columns 1 is very nice. Would it work to make the `S` a snake?

I am not very good at this, but with some help, we gave a go and here are some additional ideas to add the mix:

![Screenshot 2021-03-28 at 18 42 37](https://user-images.githubusercontent.com/11851990/112765129-9f7cf400-9003-11eb-95ed-5d901ddd13be.png)

Some of it appears not to be very good, for example, I had hope that something good will come out of putting waves/snake but it turns out just to look like a parachute...

jlaehne:
@jordiferrero, I like the typography and the simpleness of columns 1&2, with a clear preference for the black version (not only because I use the dark theme on github, but also because lumi-labs tend to be dark, which is less the case for CL though)!

1B is nice, because I think including the full name is good. On the other hand, we need a square logo for the icon. Either having 2B as a compact version or madifying the logo towards a more squared shape.

By the way, you would need to switch purple and blue ;-)

From @ericpre I like the wave idea (the parachute looks like a detector, but looks too playful for a logo probably), though the lines might have to be a bit thicker to look good. How would it look to combine 1B with the waves (still with a bit of offset to the dot of the i, which could be slightly enlarged, but not too much) - could also help going closer to a square aspect ratio of the overall logo?

jordiferrero:
I love all your comments! I will put together a version similar to 1B with waves and will share again so we can vote again!
Thanks for the involvement :-)

jordiferrero:
Here we go again...
[...]

PR #70, opened 2021-04-02
Add logo files and logo in README

jordiferrero:
### Description of the change
Finalising #65 
Added logo files (`.svg`) in the docs folder.
Added logo in the readme.

### Progress of the PR
- [x] ready for review.


ericpre:
The files you added will not be included in the package, which is not an issue as such but since one of these file is used in the `README.md`, I suspect that they will be a missing image error on pypi.
See https://packaging.python.org/guides/using-manifest-in on how to include files in distribution.

jlaehne:
I rescaled the logo in #72 and replaced path by absolute link that should also work from pypi.

PR #72, opened 2021-04-02
Resize the logo in README.md

jlaehne:
and use absolute path so that it would work even if files are not included, e.g. on pypi


ericpre:
The issue with this approach is that if the file/folder is renamed or if github changes the link, which is possible anytime, because this link is not permanent, it will be broken.
Even if we keep updating the link when necessary, the readme of older version on pypi will be broken.

Github will rename the `master` branch to the `main` branch at some point soon.

jlaehne:
OK, so I'll have another look manifest-in.

PR #78, opened 2021-06-06
Black formatting

ericpre:
### Progress of the PR
- [x] Run black on source code
- [x] add black linting github actions - see https://black.readthedocs.io/en/stable/integrations/github_actions.html
- [x] ready for review.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,11 @@
+name: Lint
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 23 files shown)


jlaehne (approved):

jlaehne:
@ericpre do you have an idea, why only the py3.8 coverage test run fails?

ericpre:
This build is using the `non_uniform_axes` branch and there is a bug somewhere, which can be either in `join_spectra` or the `non_uniform_axes` branch. It would be good to figure it out!

jlaehne:
Thanks, I did not realize that only this one test is running on the non_uniform_axes branch. Ends up in a nice catch for that branch. Fixed in https://github.com/hyperspy/hyperspy/pull/2765

PR #79, opened 2021-06-07
force r to be int

jlaehne:
### Description of the change
Due to https://github.com/hyperspy/hyperspy/pull/2765, I realized that the overlap parameter `r` in `join_spectra` could be non integer, so forcing a conversion if it should be of a different type.

### Progress of the PR
- [x] add tests,
- [x] ready for review.


jlaehne:
Once https://github.com/hyperspy/hyperspy/pull/2765 is merged, the failing test should be OK as well - we should maybe wait for that before merging this one.

ericpre (approved):

PR #84, opened 2021-08-24
Add badges

jlaehne:
Add build, anaconda and download badges and change DOI to permanent one.

### Requirements
* Read the [contributing guidelines](https://github.com/lumispy/lumispy/blob/master/.github/CONTRIBUTING.md).
* Fill out the template; it helps the review process and it is useful to summarise the PR.
* This template can be updated during the progression of the PR to summarise its status. 

*You can delete this section after you read it.*

### Description of the change
- Change A.
- Change B.
- etc.

### Progress of the PR
- [ ] Change implemented (can be split into several points),
- [ ] update docstring (if appropriate),
- [ ] add tests,
- [ ] ready for review.

```python
import lumispy as lum
import numpy as np
# Your new feature...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,16 @@
 [![Build Status](https://dev.azure.com/Lumispy/lumispy/_apis/build/status/LumiSpy.lumispy?branchName=master)](https://dev.azure.com/Lumispy/lumispy/_build/latest?definitionId=3&branchName=master)
 ![Tests](https://github.com/lumispy/lumispy/workflows/Tests/badge.svg)
+[![Build status](https://github.com/pyxem/kikuchipy/workflows/build/badge.svg)](https://github.com/pyxem/kikuchipy/actions)
 [![Codecov Status](https://codecov.io/gh/lumispy/lumispy/branch/master/graph/badge.svg?token=8ZFX8X4Z1I)](https://codecov.io/gh/lumispy/lumispy)
-
 [![License: GPL v3](https://img.shields.io/badge/License-GPLv3-blue.svg)](https://www.gnu.org/licenses/gpl-3.0)
+
 [![Python Version](https://img.shields.io/pypi/pyversions/lumispy.svg?style=flat)](https://pypi.python.org/pypi/lumispy)
 [![PyPi Version](http://img.shields.io/pypi/v/lumispy.svg?style=flat)](https://pypi.python.org/pypi/lumispy)
-[![DOI](https://zenodo.org/badge/222482347.svg)](https://zenodo.org/badge/latestdoi/222482347)
+[![PyPi Downloads](https://static.pepy.tech/personalized-badge/lumispy?period=total&units=international_system&left_color=grey&right_color=orange&left_text=downloads)](https://pepy.tech/project/lumispy)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,11 +1,16 @@
 [![Build Status](https://dev.azure.com/Lumispy/lumispy/_apis/build/status/LumiSpy.lumispy?branchName=master)](https://dev.azure.com/Lumispy/lumispy/_build/latest?definitionId=3&branchName=master)
 ![Tests](https://github.com/lumispy/lumispy/workflows/Tests/badge.svg)
+[![Build status](https://github.com/pyxem/kikuchipy/workflows/build/badge.svg)](https://github.com/pyxem/kikuchipy/actions)
```
ericpre, line 3: It seems that the copy and paste is missing some update! :)
This badge is the same as the "Tests" badge and the previous one is missing the link to the actions tab..

README.md
```diff
@@ -1,11 +1,16 @@
...
 [![Codecov Status](https://codecov.io/gh/lumispy/lumispy/branch/master/graph/badge.svg?token=8ZFX8X4Z1I)](https://codecov.io/gh/lumispy/lumispy)
-
 [![License: GPL v3](https://img.shields.io/badge/License-GPLv3-blue.svg)](https://www.gnu.org/licenses/gpl-3.0)
+
 [![Python Version](https://img.shields.io/pypi/pyversions/lumispy.svg?style=flat)](https://pypi.python.org/pypi/lumispy)
 [![PyPi Version](http://img.shields.io/pypi/v/lumispy.svg?style=flat)](https://pypi.python.org/pypi/lumispy)
-[![DOI](https://zenodo.org/badge/222482347.svg)](https://zenodo.org/badge/latestdoi/222482347)
+[![PyPi Downloads](https://static.pepy.tech/personalized-badge/lumispy?period=total&units=international_system&left_color=grey&right_color=orange&left_text=downloads)](https://pepy.tech/project/lumispy)
+[![Anaconda Version](https://anaconda.org/conda-forge/lumispy/badges/version.svg)](https://anaconda.org/conda-forge/lumispy)
```
ericpre, line 10: Anaconda download are known to largely over estimated, not only because it can be used by CI, but there is also some download of mirror.

pypi downloads are also known to be inaccurate, but I don't remember the reason. At some point it was discussed in the hyperspy repo.

README.md
jlaehne: Indeed, and the workflow is actually called tests in our case and already included. I was a bit quick on that one.

README.md
jlaehne: That is what I guessed as well when seeing the numbers, I had just looked what type of bdsges kikuchipy and pyxem have. So maybe we should leave them out, as it is not very instructive numbers.

ericpre: Yes, these numbers are misleading but you will see that there are used fairly widely and wrongly used obviously... It is briefly mentioned in the hyperspy developer guide: https://hyperspy.org/hyperspy-doc/current/dev_guide/useful_information.html#monitoring-version-distribution

More accurate numbers would be through google analytics of a documentation website, but it also has its shortcomings! The number of stars of the github repository can also be used to estimate how widely is a library used.

PR #94, opened 2021-11-23
Make links work on PyPi

jlaehne:
Internal GitHub links do not work for `README.md` on PyPi (https://pypi.org/project/lumispy/), so replace by global links.


jordiferrero:
Thanks

PR #98, opened 2022-01-17
Jacobian transformation of the variance model

### LINKED ISSUE #96: Setting a variance model in the lumispy object after Jacobian transform
After some discussion with @LMSC-NTappy, we realised that the variance model stored in the `LumiSpectrum.metadata.Signal.Noise_properties.variance` should no longer be homoscedastic.

Instead, the noise variance should also be converted to eV, but with the intensities multiplied by the square of the jacobian (according to `Var(aX) = a**2*Var(X)`).

I am not sure what the best way to implement this subtlety is in our luminescence object. I understand that, once you have your signal, then you can compute the variance manually and then apply the Jacobian transformation. But is there a way to define this noise in a generic way (to the class rather than to the data), so when the `cl.to_eV()` function is called, such transformation is also applied to the variance metadata (should there be any).

[This](http://hyperspy.org/hyperspy-doc/current/user_guide/signal.html?highlight=noise#setting-the-noise-properties) is the Hyperspy guideline on noise metadata.

Let's use this chat to discuss further @LMSC-NTappy. Any thoughts?

jlaehne:
### Description of the change
As brought up in #96, the Jacobian transformation of the data implies renormalization of the variance by the squared Jacobian, which is implemented in this PR for the `to_eV` and `to_invcm` methods.

Closes #96.

### Progress of the PR
- [x] handle variance if Jacobian is not performed,
- [x] implement for invcm_relative (check if jacobian is properly implemented there),
- [x] ensure proper formatting,
- [x] add tests,
- [X] ready for review.

```python
import lumispy as lum
import numpy as np
s.estimate_poissonian_noise_variance()
s.to_eV()
```


jlaehne:
@jordiferrero @LMSC-NTappy the base functionality should be in this PR, maybe you can already do a check whether it is doing what you would expect.

jlaehne:
For completeness, we should also handle the variance when the Jacobian is not performed (as we have the option). I guess, in that case the data would just be flipped and the axis exchanges for the variance signal - just like for the main signal?

lumispy/signals/luminescence_spectrum.py
```diff
@@ -107,10 +111,20 @@ def to_eV(self, inplace=True, jacobian=True):
                     self.axes_manager.signal_axes[0],
                     evaxis.axis,
                 )
+                if self.metadata.has_item("Signal.Noise_properties.variance"):
```
LMSC-NTappy, line 114: Here, I wonder what happens in case variance is set using the `noise_properties.variance_linear_model` and not the signal object

jordiferrero: I suppose we should first run the `s.estimate_poissonian_noise_variance()` which would return a signal object for the metadata. Then the function would proceed correctly. Am I correct? Maybe worth including this case too.

jordiferrero (commented):
Thanks for pulling this off so quickly. Since it's Nicolas' work mainly, feel free to merge when he is happy with the changes :-)

lumispy/signals/luminescence_spectrum.py
```diff
@@ -194,45 +203,43 @@ def to_invcm(self, inplace=True, jacobian=True):
...
-                        self.axes_manager.navigation_axes[0].get_axis_dictionary(),
-                        invcmaxis.get_axis_dictionary(),
-                    ),
-                )
-            s2.set_signal_type(self.metadata.Signal.signal_type)
-            s2.metadata = self.metadata
+
+            s2 = self._deepcopy_with_new_data(s2data)
+            s2.axes_manager.remove(-1)
+            s2.axes_manager._axes.append(invcmaxis)
```
LMSC-NTappy, line 234: Same remark as above

jlaehne:
Well this was definitely just a start. I wanted to get started with the specific use case you described. We'll have to put in more thought to adress all use cases and I'm happy for your input as you seem to have dived deeper in the variance handling in HyperSpy already. From what I see at the moment, we have to cover the following cases:

- If no variance is set, we do not need to initiate it and should just run the conversion without - that is why we should not simply run `s.estimate_poissonian_noise_variance()` in any case.
- The variance can just be a constant (homoscedastic noise). I guess in that case we would first have to create a noise signal object that has the same dimensions as the data, but filled with a constant value everywhere? After the conversion we would have heteroscedastic noise.
- The variance can be set as a signal (heteroscedastic noise, either using `estimate_poissonian_noise_variance()` or defined as the user likes) -- that is the case handled so far, the signal object is just taken as it is.
- As far as I understand, the `variance_linear_model` is not independent, but used in `estimate_poissonian_noise_variance()`. So in that case it would also alter the expected values that we set after the transformation. The question is then, do the constants defining this model need to be adapted in the course of the transformation?

In case the Jacobian is not performed, a constant variance would be kept, while for heteroscedastic noise, the axes needs to still be replaced and the data order flipped.

> Is there a reason you use the DataAxis formalism for the ev signal and not FunctionalDataAxis? It looks to me that this would simplify part of the procedure.

Yes, the reason is that the conversion to energy axis is implemented taking into account the wavelength-dependent permittivity of air defined in `_n_air()`. The correction is not that big (0.4 meV at 200 nm), but for completeness I had implemented that some years ago. However, that formula makes it impractical to use `FunctionalDataAxis`.

LMSC-NTappy:
>the reason is that the conversion to energy axis is implemented taking into account the wavelength-dependent permittivity of air >defined in _n_air()

Fair enough. That's some serious completeness right here! 0.4 meV does mean something in some cases though... This makes me realize that sooner than later we'll need to define axes objects dedicated to spectroscopy. 

>As far as I understand, the variance_linear_model is not independent, but used in estimate_poissonian_noise_variance(). So in that case it would also alter the expected values that we set after the transformation. The question is then, do the constants defining this model need to be adapted in the course of the transformation?

Short answer: yes it has to be adapted but it's a bit complicated because we cannot define the way it needs renormalisation by just adapting the floating-point values (the gain-offset should be multiplied by jacobian-squared which is an array). Ideally I'd like this to change in hyperspy directly, also because imho the current way this noise model is defined is a bit hard to understand and unpractical... A time will come but not today :)
Anyways, for this reason I recommend applying the already-defined noise parameters of the model, if they exist, and then reset them explicitly, while raising a warning that it is a normal consquence: 
'signal.estimate_noise_variance(expected_value=variance_signal,gain_factor=1.0,gain_offset=0.0,correlation_factor=1.0)'

I 100% agree with you otherwise. Final remark, in my understanding is that in all cases, at use time (i.e. when the fit is performed) the variance is broadcasted into a np.array of the same shape as `signal.data` so that would be the way to deal with the existing variance signal if `signal.get_noise_variance()` returns a float (I also don't like that it is allowed to return either BaseSignal or float btw ^^).

jlaehne:
Just as an update, I will probably only find time to finish this in about two weeks.

jlaehne:
The different cases described above our now handled:
- variance signal
- constant variance
- variance not set
for 
- jacobian=True/False
- inplace=True/False

`variance_linear_model` is reset to defaults if jacobian=True, otherwise kept
constant variance is converted to signal if jacobian=True

The only thing, I have not implemented yet is running an `estimate_poissonian_noise_variance` prior to the transformation. We could either do so in every case that `variance_linear_model` deviates from the defaults? Or write in the documentation that it has to be run manually in case the user wants to be sure that the parameters are taken into account. I tend to the former, as I also do not run it for the other situations, but work with the variance object as it is,. I believe that is the more flexible solution. 

What do you think @LMSC-NTappy @jordiferrero - in case you have data files that one could test the procedures with, I would be grateful if you could test the functionality and give me feedback.

jordiferrero (approved):

jordiferrero:
To me it looks great!
I would suggest @LMSC-NTappy to do a final check and merge.
Let's just make sure the documentation is clear on what it can and cannot do (to date).

Thanks a lot (again)!

jlaehne:
> Let's just make sure the documentation is clear on what it can and cannot do (to date).

Yes, as soon as the basic points for the docs are sorted out and this PR is merged, I would go about writing a proper documentation for the energy conversion and the Jacobian transformation including the operational principle on the variance transformation. @LMSC-NTappy could then maybe contribute a bit more on the background of the variance topic.

LMSC-NTappy:
>  @LMSC-NTappy could then maybe contribute a bit more on the background of the variance topic.

Sure with pleasure!

I also think it could be merged. I think the tests could be a bit more oriented towards the results of the operations themselves, but in the end they make sure the code is run so that's already a large part of the goal. The rest is up to you

Cheers

Nico

jlaehne:
Great, I'll merge it.

> I think the tests could be a bit more oriented towards the results of the operations themselves
@LMSC-NTappy would you be able to contribute a few additional such tests to make sure that it is not only formally correct, but also scientifically?

LMSC-NTappy:
@jlaehne I sure can. 

Note that I am sure that the science is correct =).

The goal is more to provide a reference point for future changes in the code. E.g. dealing with functionaldataaxes.

Issue #100, opened 2022-02-05
Added Read the Docs support


jordiferrero:
Hi all,

I have wanted to do this for very long...
I have now added support for the read the docs. This means that we can have explanations on the functions we write/methodologies in a nice formatted way.

As far as I am aware, such guides can be added to `user_guide/file.rst`. Feel free to add content there.
I have a list of pages I would like to add for the analysis of PL data (coming soon).
@jlaehne Maybe you could take care on the [Jacobian transform documentation](https://lumispy.readthedocs.io/en/latest/user_guide/signal_axis.html) if you feel like explaining in a bit more detail how LumiSpy does this transform. That'd be a great opportunity to test it out too.

I believe having this will help external users to get started, as a nicely explained guide (like the EDX analysis page in HyperSpy docs) can really become "the place to check" on how to do data analysis of a particular data signal. I think the demo notebooks are great, but they can also get a bit messy.

Happy to hear suggestions!

PS: The Read the docs should trigger a build every time there is a merge. It will loop though all the docstrings and update the docs automatically.

jordiferrero:
This issue can be closed anytime, as it has already been resolved. It was more of an announcement.

jlaehne:
Great, thanks @jordiferrero for starting out on this. I have been wanting to suggest this for a while as well. Now that the basic setup is done, it should be straightforward to contribute.

I would propose that I redirect the domain https://lumispy.org to the documentation then.

Given our logo, we should maybe use a dark theme as standard. For example that is possible with the theme used by https://kikuchipy.org (though the background behind the logo is not completely black there).

jlaehne:
> Given our logo, we should maybe use a dark theme as standard. For example that is possible with the theme used by https://kikuchipy.org (though the background behind the logo is not completely black there).

See #103

jlaehne:
ToDo:
- [x] Adapt color behind logo, provide dark theme (see #103),
- [ ] Write introductory page that gives an overview of the signal types under development and summarizes functionality present and planned,
- [ ] Write documentation about energy/wavenumber conversion and Jacobian,
- [ ] Add documentation on variance,
- [ ] Write documentation on fitting,
- [ ] Give overview of other utility functions,
- [ ] Merge Installation guide in the main folder into the new one in the documentation and correct links, 
- [ ] Mention license somewhere in the documentation,
- [ ] Link realeasing_guide (convert to .rst) from the contributor's instructions,
[...]

PR #103, opened 2022-02-14
add dark theme support

jlaehne:
Adds dark theme support based on the pillow theme: https://pillow.readthedocs.io/en/stable/

Logo background is set to black for both dark and light themes.

Some first additions to the documentation, incorporate existing pages in the main folder (changelog, contributing guide).


jlaehne:
See https://lumispy--103.org.readthedocs.build/en/103/citing.html to check out the docs. In Firefox use `Ctrl-Shift-I` to activate the development tools. Under 'Inspector' you can switch between dark (moon icon) and light (sun icon) modes.

jordiferrero (approved):

jordiferrero:
Thanks for this.
Do we need to include the changelog in the documentation so visible? I followed the hyperspy documentation and did not see it there. I would consider removing it.

jlaehne:
> Thanks for this. Do we need to include the changelog in the documentation so visible? I followed the hyperspy documentation and did not see it there. I would consider removing it.

I would like to have it in the docs, but it does not need to be so prominent.

In HyperSpy, it is the second point in the User Guide after introduction: https://hyperspy.org/hyperspy-doc/current/index.html

I could also see it in the `API References` section instead of the user guide? What do you think?

> Otherwise, feel free to merge :-)
I'll wait for the decision on .rst vs .md - in case we opt for .md I would revert the files that I have converted back to .md to avoid a back and forth, but leave the rest of the conversion for a separate PR.

ericpre:
As a library user, I find comprehensive change log very useful and important!

.github/CONTRIBUTING.md
```diff
@@ -1,9 +0,0 @@
-## Issues
```
ericpre, line 1: This file is used by github - see https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors and https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/encouraging-helpful-contributions-to-your-project-with-labels.

From a quick at the github documentation, rst is not mentioned anywhere, do we know if github can parse the `rst` format? Maybe simply add a link to a markdown file located at the right place, as there are alternatives. Using the https://github.com/LumiSpy/.github would have the advantage of keeping this repository tidy.

ericpre (approved):

.github/CONTRIBUTING.md
jlaehne: Should be doable with the Markdown extension for Sphinx: https://www.sphinx-doc.org/en/master/usage/markdown.html

Then the files in the main folder would remain .md and could still be integrated in the docs. I'll check that out later today.

jlaehne:
> As a library user, I find comprehensive change log very useful and important!

Any thought on putting it in the `API references` section @ericpre?

.github/CONTRIBUTING.md
jlaehne: But, `rst` would be supported by GitHub as well: https://docs.github.com/en/repositories/working-with-files/using-files/working-with-non-code-files#rendering-differences-in-prose-documents

ericpre:
I wouldn't put it in the API reference. In my opinion, the best location will be at the same level than `User guide`, `API reference`, etc., as this is for done for [scipy](https://scipy.github.io/devdocs/index.html).

On a related topic, I really like the [pydata-sphinx-theme](https://pydata-sphinx-theme.readthedocs.io/en/latest/), because I find it very clear and functional.

.github/CONTRIBUTING.md
ericpre: OK, that would be brilliant! Then, once this is merged, github should show us a notification when making a new PR, because it should detect that the contributing guide has changed.

jlaehne:
Not so easy to integrate `.md` as I thought, so only the files not imported to the docs remain in markdown.

Already started out on getting docstrings corrected so that they are properly parsed in the api documentation, but there are still a few issues remaining (added it to the todo in #100).

I'll merge now so that we can continue in new PRs.

PR #106, opened 2022-02-17
Add Toggleprompt to sphinx

jlaehne:
Allows to toggle `>>>` in the examples in the documentation. Adapted from HyperSpy.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -18,3 +18,5 @@ python:
    install:
       - method: pip
         path: .
+        extra_requirements:
+           - build-doc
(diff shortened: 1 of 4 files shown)


jlaehne:
@ericpre Do you have an idea what else needs to be set up for the doc build to run through with `sphinx_toggleprompt`? Do I need to change something in the configuration on the read-the-docs site?

ericpre:
I don't know really but the `readthedocs.yaml` configuration file needs to be updated to install required dependencies, see for example: https://github.com/hyperspy/hyperspy/blob/3663b930bbee6c7f6b1933a807514007a88a3334/readthedocs.yaml#L21-L22

jlaehne:
Perfect, thanks @ericpre for the quick response. That were exactly the missing two lines.

Issue #53, opened 2021-02-21
Luminescence specific metadata


jordiferrero:
I have recently tried to think of what parameters we should include in the `s.metadata` and what parameters stay as `.original_metadata` (and  #51 made me start this issue).

I normally load `.sur` files using the hyperspy io plugin. That reads the orignal_metadata (more than 50 parameters) and then I parse some relevant values to the metadata to be used in some functions (e.g. calibration, normalisation...).

- What parameters should those be? They would add up to the hyperspy specific metadata structure.
- How would we parse metadata in an automatic way (so from the original_metadata) depending on the initial file extension loaded?

jlaehne:
I recently contributed a selective inclusion of CL metadata from Gatan files to HyperSpy: https://github.com/hyperspy/hyperspy/pull/2590

Gatan, unfortunately has a very different metadata structure, depending on acquisition mode (PMT/CCD) and again for spectral images. I tried to include the most important fields in `metadata.Acquisition_instrument.CL`

However, so far it does not include post-processing done in Digital Micrograph. Which eventually would be good in view of #51. Though, multiple post-processing steps lead to a recursive metadata structure, which would have to be properly parsed.

As it might not always be possible to enforce the same metadata hierarchy for all instrument types (though desirable), I contributed the metadata search capability in hyperspy/hyperspy#2633

jordiferrero:
That's very nice!
I guess my one comment is: should we have the CL specific metadata structure to be exclusive to lumispy?

I feel that the metadata search capability is definitely something for hyperspy, but maybe the CL metadata fields should be part of lumispy only?

What are your thoughts?

jlaehne:
The DigitalMicrograph reader was already parsing metadata for EELS and EDS, so I thought it makes sense to add the functionality for CL metadata there as well.

jlaehne:
The longterm aim for HyperSpy is that all the IO functions will move to an extension package at some point.

For me it makes more sense to handle the metadata when the file is loaded, otherwise one has to run an extra function on the data, or we again need a wrapper for the load function, no?

However, I do not know how it is for the reader that you use, could the metadata parsing be added there?

Despite the search functionality, we should in any case try to stick to some basic conventions where to save the essential metadata in the tree.

jordiferrero:
The loader I use is for `.sur` files mainly.
That one parses all the `s.original_metadata` very nicely.
[...]

PR #109, opened 2022-02-22
Metadata structure

jlaehne:
Introduce LumiSpy specific conventions on metadata structure as brought up in #53.

Decisions I have taken for the moment:
- 'Spectrometer', 'Detector', and 'Spectral_image' are independent nodes on the 'Acquisition_instrument' level. None of them is really hierarchically below any of the others.
- 'Detector' is the general node and contains the 'detector_type'. This way, general parameters such as the 'exposure_time' are not hidden in a detector-specific subnode.
- Besides 'Grating', the 'Spectrometer' contains an additional subnode 'Filters'.

Open questions:
- Should we have separate 'entrance_slit_width' and 'exit_slit_width' parameters? In serial acquisition both can play a role and (though uncommon) may theoretically be set independently. Alternative would be have a general 'slit_width' leaf.
- The biggest issue is the different conventions on the integration time. Both CL and PL systems I work with allow for multiple 'frames' with a certain 'dwell_time' that add up to an overall 'integration_time'. The attolight system has a more specific distinction between 'real_time' and 'life_time'. How does that fit in, which one should be the integration_time? We should have one commonly defined parameter complemented by additional ones.
- Overall, this is just a convention on the structure and proposal for major leafs that can be contained. Everyone is free to add additional information or leave out some. As such the structure is somewhat exemplary - so we could out leave more specific leaves like 'saturation_fraction', 'signal_amplification', 'readout_rate' from the documentation.
- Where would we put the measurement temperature?

Broader questions:
- Similar to detector, a 'Excitation_source' could make sense. Alternatively, the 'SEM/TEM' nodes of HyperSpy could be complemented by a 'Laser' node.

### DIFF
--- a/doc/source/user_guide/metadata_structure.rst
+++ b/doc/source/user_guide/metadata_structure.rst
@@ -59,7 +59,7 @@ while parallel acquisition with a CCD is characterized by the
         │   ├── signal_amplification (xN)
         │   ├── readout_rate (MHz)
         │   └── pixel_width (mm)
-        └── SpectralImage
+        └── Spectral_image
             ├── mode
             ├── step_size
             ├── step_size_units (nm)
@@ -236,7 +236,7 @@ pixel_width
     Diameter of a pixel in mm.
 
 
-SpectralImage
+Spectral_image
 =============
 
 Contains information about mapping parameters, such as step size, drift
--- a/doc/source/user_guide/metadata_structure.rst
+++ b/doc/source/user_guide/metadata_structure.rst
@@ -16,6 +16,14 @@ the ones for leaves are not capitalized. When a leaf contains a quantity that
 is not dimensionless, the units can be given in an extra leaf with the same
 label followed by the ``_units`` suffix.
 
+Besides directly accesing the metadata tree structure, e.g.
+``s.metadata.Signal.signal_type``, the HyperSpy methods
+:external:py:meth:`hyperspy.misc.utils.DictionaryTreeBrowser.set_item`,
+:external:py:meth:`hyperspy.misc.utils.DictionaryTreeBrowser.has_item` and
+:external:py:meth:`hyperspy.misc.utils.DictionaryTreeBrowser.get_item`
+can be used to add to, search for and read from items in the metadata tree,
+respectively.
+
 The luminescence specific metadata structure is represented in the following
 tree diagram. The default units are given in parentheses. Details about the
 leaves can be found in the following sections of this chapter. Note that not
@@ -26,6 +34,7 @@ while parallel acquisition with a CCD is characterized by the
 
 ::
 
+    metadata
     ├── General
     │   └── # see HyperSpy
     ├── Sample
--- a/doc/source/user_guide/metadata_structure.rst
+++ b/doc/source/user_guide/metadata_structure.rst
@@ -18,7 +18,7 @@ label followed by the ``_units`` suffix.
 The luminescence specific metadata structure is represented in the following
 tree diagram. The default units are given in parentheses. Details about the
 leaves can be found in the following sections of this chapter. Note that not
-all types of leaves will apply to every type of measurements. For example,
+all types of leaves will apply to every type of measurement. For example,
 while parallel acquisition with a CCD is characterized by the
 ``central_wavelength``, a serial acquisition with a PMT will require a
 ``start_wavelength`` and a ``step_size``.
@@ -26,8 +26,19 @@ while parallel acquisition with a CCD is characterized by the
 ::
 
     └── Acquisition_instrument
+        ├── General
+        │   └── see HyperSpy
+        ├── Sample
+        │   └── see HyperSpy
+        ├── Laser / SEM / TEM
+        │   ├── laser_type
+        │   ├── model
+        │   ├── wavelength (nm)
+        │   ├── power (mW)
+        │   ├── objective_magnification
+        │   └── for SEM/TEM see HyperSpy
         ├── Spectrometer
-        │   ├── model_name
+        │   ├── model
         │   ├── acquisition_mode
         │   ├── entrance_slit_width (mm)
         │   ├── exit_slit_width (mm)
@@ -46,7 +57,7 @@ while parallel acquisition with a CCD is characterized by the
         │       └── cut_off_wavelength (nm)
         ├── Detector
         │   ├── detector_type
-        │   ├── model_name
+        │   ├── model
         │   ├── frames
         │   ├── integration_time (s)
         │   ├── saturation_fraction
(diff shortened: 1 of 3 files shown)


jlaehne:
I would like to decide on the main structure rather soon, so that I can adapt the handling for digital micrograph from https://github.com/hyperspy/hyperspy/pull/2590 before it is released in [v1.7](https://github.com/hyperspy/hyperspy/milestone/40).

You can preview the metadata documentation at: https://lumispy--109.org.readthedocs.build/en/109/user_guide/metadata_structure.html

@jordiferrero @LMSC-NTappy @dnjohnstone @mikelgr11

LMSC-NTappy:
Looks good to me. I think it encompasses pretty well what happens in the experiment in a very compact definition. 

Regarding your open questions:

>Should we have separate 'entrance_slit_width' and 'exit_slit_width' parameters? In serial acquisition both can play a role and (though uncommon) may theoretically be set independently. Alternative would be have a general 'slit_width' leaf.

I would definitely add them both, also in anticipation of the multi-spectrometer setups, e.g. for Raman spectroscopy

>The biggest issue is the different conventions on the integration time. Both CL and PL systems I work with allow for multiple 'frames' with a certain 'dwell_time' that add up to an overall 'integration_time'. The attolight system has a more specific distinction between 'real_time' and 'life_time'. How does that fit in, which one should be the integration_time? We should have one commonly defined parameter complemented by additional ones.

I think this is mostly different based on the camera used for the system, and whether the sensor performs on-chip frame storage (aka Kinetics mode and variants in PI or Andor CCDs). In this context, I think life_time refers to the actual time the sensor is exposed to light and real_time the time it takes to perform the digitization operation, since the readout is decoupled. I would use a single integration time and number of frames fields, since the rest is probably in the `original_metadata` anyways.

>Similar to detector, a 'Excitation_source' could make sense. Alternatively, the 'SEM/TEM' nodes of HyperSpy could be complemented by a 'Laser' node.

Yes most certainly.

jlaehne:
> Generally I am in favor of keeping this simple, and handle the details of the metadata in the `original_metadata` of the signal object. The goal is to avoid over-interpreting the metadata: for example Readout_rate (MHz) probably refers to the ADC readout rate since it is in MHz, but depending on the manufacturer it could also mean the frame rate (number of readouts per second). So that's a point where documentation is crucial I think

Yes, I agree. We should not overload this part. I guess `signal_amplification (xN)` could also remain in the `original_metadata`. Same with `saturation_fraction`. Anyway, the metadata can always contain also fields that we do not include in the documentation here, but which are not necessarily available for any manufacturer.

LMSC-NTappy:
The only fear that I have with signal amplification is that it might be used for both the ADC sensitivity and the EM-Gain for EMCCDs which are two very different beasts

jlaehne:
Implemented the suggestions of @LMSC-NTappy

Also removed the additional node `Filters`, filters should just be numbered sequentially if more than one is present (same with spectrometers), I would propose.

jlaehne:
Implemented for digital micrograph in https://github.com/hyperspy/hyperspy/pull/2894

I'm waiting for feedback from one more colleague and then I would merge.

doc/source/user_guide/metadata_structure.rst
```diff
@@ -0,0 +1,361 @@
...
+binning
+    type: tuple of int
+
+    A tuple that describes the binning of a parallel detector such a CCD on
+    readout in x and y directions.
+
+processing
+    type: string
+
+    Information about automatic processing performed on the data, e.g. 'dark
```
ericpre, line 318: If this is going to be used by lumispy, for example, to know if a processing needs to be applied, would it be good to standardise it more tightly by allowing only a set of specific strings?

doc/source/user_guide/metadata_structure.rst
```diff
@@ -0,0 +1,361 @@
...
+
+sensor_roi
+    type: tuple of int
+
+    Tuple that specifies range of pixels on a detector that are read out.
+
+pixel_width
+    type: float
+
+    Diameter of a pixel in µm.
```
ericpre, line 329:
```suggestion
pixel_width
    type: float or tuple of float

    Size of a pixel in µm. Tuple of length 2 (width, height), when the pixel is not square.
```

doc/source/user_guide/metadata_structure.rst
```diff
@@ -0,0 +1,361 @@
...
+processing
+    type: string
+
+    Information about automatic processing performed on the data, e.g. 'dark
+    subtracted'.
+
+sensor_roi
+    type: tuple of int
+
+    Tuple that specifies range of pixels on a detector that are read out.
```
ericpre, line 324: Specify the order of the values of the tuple? Which one is position, size, x, y, etc.?

doc/source/user_guide/metadata_structure.rst
```diff
@@ -0,0 +1,361 @@
...
+    type: float
+
+    Emission wavelength of the exciting laser in nm.
+
+power
+    type: float
+
+    Measured power of the excitation laser in mW.
+
+objective_magnification
```
ericpre, line 162: Would `magnification` on its own be enough?

doc/source/user_guide/metadata_structure.rst
```diff
@@ -0,0 +1,361 @@
...
+
+step_size
+    type: float
+
+    Distance between subsequent pixels in the spectral image.
+
+step_size_units
+    type: string
+
+    Units of the step size (standard 'nm').
```
ericpre, line 351: Is not redundant with the information provided in the `AxesManager`?

doc/source/user_guide/metadata_structure.rst
jlaehne: Well, the laser itself is not magnified, that is why I came up with this additional specification. On the other hand, SEM/TEM do have a field `magnification` so it could make sense to use the same here.

doc/source/user_guide/metadata_structure.rst
jlaehne: For the moment, I do not have an overview what different types of information and strings we could get from vendor software. For the moment, I would leave it.

doc/source/user_guide/metadata_structure.rst
jlaehne: Good point, we should maybe name it pixel_size then.

doc/source/user_guide/metadata_structure.rst
jlaehne: Indeed, it is in the definition of the navigation axes, so we can leave it out here.

ericpre (approved):

jlaehne:
Thanks.

jordiferrero:
Thank you for that. I agree that it is simple enough.

@LMSC-NTappy, what would be the easiest way to parse some of the original metadata from the `.sur` files to this new metadata structure? During the `CLSEMSpectrum` class intialization?

jlaehne:
As addendum, I just have proposed to the FAIRmat consortium (that creates data infrastructure for materials science data in Germany), which is currently working on Nexus file format definitions for additional data types, to create a Nexus class for luminescence spectroscopy based on the metadata format we defined here. Thus, there might be some feedback in the next weeks from other groups, but in the end we would have a metadata format compatible with the one in the corresponding Nexus definition.

LMSC-NTappy:
Hi @jordiferrero, @jlaehne,

All the metadata from the ".sur" file is stored in the `.original_metadata` field of the hyperspy object. When the `CLSEMSpectrum` object is instanciated, it can already access it and pick inside whatever it needs.

I think the only caveat is that it's passed as a dictionary, so you need to access, e.g.
`signal_dict['original_metadata']['Object_0_Channel_0']['Parsed']['CCD']['Exposure_Time']`

jlaehne:
We started discussing how to best implement a parser for metadata in #53. We should probably implement such a general metadata framework for the HyperSpy universe.

LMSC-NTappy:
I agree.

Then best I can think of is that it could be implemented as a `'post_process'` function in the loader. See:
https://github.com/hyperspy/hyperspy/blob/59fe2c242052aab593b8b4b6beb62cdd227fda5a/hyperspy/io.py#L721-L723

The post_process function can be defined from the loader AND access signal-methods since the instantiated signal is passed to it as an argument. So it has potential to interface extensions with main hyperspy. However I doubt we can decouple both completely.

jlaehne:
Lets continue discussing in #53, there were more people involved already there.

PR #114, opened 2022-03-18
Fixing deprecation warning

### LINKED ISSUE #54: Test deprecation warning
#44 introduced the following deprecation warning, which we should fix:
```
lumispy/tests/test_luminescence_spectrum.py::TestLumiSpectrum::test_errors_raise
  /home/jonas/medien/pdi/git/lumispy/lumispy/signals/luminescence_spectrum.py:312: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
    background_xy = np.array(background)
```

jordiferrero:
### Description of the change
Try to fix depreciation warning #54 .
Depreciate the `remove_background_from_file` method. It will be replaced by a new better function using the HyperSpy syntaxis.

### Progress of the PR
- [x] ready for review.


jordiferrero:
I believe the DepreciationWarning caused by luminescence_spectrum.py has been fixed.
There is still the following warning:
`lumispy/tests/signals/test_luminescence_spectrum.py::TestLumiSpectrum::test_errors_raise
  /opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/numpy/core/fromnumeric.py:1970: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
    result = asarray(a).shape
`
which seems to come from numpy, not us. Is that correct?

jlaehne (approved):

ericpre (commented):
Looking at CI, the warning is still there. This PR slightly improves the syntax and make this function more readable.

It looks to me, that this method is not very clear; it would benefit from rewriting and possibly changing its behaviour:
- the name `remove_background_from_file` implies that we would provide a file, but the docstring only mention arrays or Signal1D for the `background` argument
- the flexibility in supporting different shape for the `background` argument is confusing.

As alternative to the current behaviour, I would suggest to support only `Signal1D` and add a separate method to read the background from a file, which returns a Signal1D. This would have the advantage of being transparent/explicit for the users. 

> I believe the DepreciationWarning caused by luminescence_spectrum.py has been fixed. There is still the following warning:
> 
> ```
> lumispy/tests/signals/test_luminescence_spectrum.py::TestLumiSpectrum::test_errors_raise
>   /opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/numpy/core/fromnumeric.py:1970: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
>     result = asarray(a).shape
> ```
> 
> which seems to come from numpy, not us. Is that correct?

It most likely means that we are not using it as it is now expected to be used and there is some changes required in lumispy to remove this warning.

The description in https://github.com/LumiSpy/lumispy/pull/114#issue-1173787255 doesn't match with what this PR does! ;)

jlaehne:
> It looks to me, that this method is not very clear; it would benefit from rewriting and possibly changing its behaviour:

Indeed, creating a variable `background` that contains the x and y vectors and afterwards splitting it again is not very intuitive (and avoiding that would along the way solve the problem with the deprecation warning.

I agree with @ericpre that the "HyperSpy" way should be to provide a method where a `Signal1D` object is subtracted from another Signal1D object. That would make it much more valuable for others to reuse. If the axes don't match (in size, offset and scale for UniformDataAxis), the background signal should be rebinned/interpolated to the main signal axis. This would also ease an extension to work in non-uniform data axes in the future.

I would propose in fact writing a new method `remove_background_signal` to do the job, and adding a deprecation warning to this method in order not "surprise" anyone with the change.

ericpre:
> I would propose in fact writing a new method `remove_background_signal` to do the job, and adding a deprecation warning to this method in order not "surprise" anyone with the change.

Other alternative would be:
- rename `remove_background_from_file` `remove_background` to support the `background` argument and use `remove_background` of `Signal1D` when `background`, as it is currently doing. This pattern is used for various method, such `plot`, `decomposition`, etc.
- implement it directly in the `remove_background` of `Signal1D` - thinking once more about, maybe the best way would be to add support for fitting a [`ScalableFixedPattern`](https://hyperspy.readthedocs.io/en/latest/api/hyperspy._components.scalable_fixed_pattern.html#hyperspy._components.scalable_fixed_pattern.ScalableFixedPattern), as it would fit better in the flow of `Signal1D.remove_background` and this would allow to compare different background using the GUI.

jlaehne:
As this was originally intended as a quick fix for some warnings, I would propose the following way forward

- already add a deprecation warning to the existing function
- create an issue that the improved functionality needs to be implemented (looks like the idea with the `ScalableFixedPattern` sounds most promising)
- then merge this PR

ericpre:
@jordiferrero, do you think, this is feasible to finish this PR for the 0.2 release, i. e. in the coming days?

lumispy/signals/luminescence_spectrum.py
```diff
@@ -512,6 +512,8 @@ def remove_background_from_file(self, background=None, inplace=False, **kwargs):
         -----
         This function does not work with non-uniform axes.
         """
+        warnings.warn('The use of `remove_background_from_file` is depreciated. '
+                      'Please use `remove_backgroun` from Signal1D object', DeprecationWarning, stacklevel=2)
```
jlaehne:
```suggestion
        warnings.warn('The use of `remove_background_from_file` is deprecated and will be removed in LumiSpy 1.0. '
                      'Please use `remove_background` from the Signal1D class.', DeprecationWarning, stacklevel=2)
```

jlaehne (approved):

jordiferrero:
@jlaehne Not it is ready for review. Can you maybe help me tell how to fix the "linting" error? That's something new to me :-)

lumispy/tests/signals/test_luminescence_spectrum.py
```diff
@@ -71,3 +71,9 @@ def test_warnings(self):
         with pytest.warns(SyntaxWarning) as warninfo:
             s.remove_background_from_file(background=None, display=False)
         assert warninfo[0].message.args[0][:18] == "Using the Hyperspy"
+
+    def test_depreciation_warning(self):
+        s = LumiSpectrum(np.ones(50))
+        with pytest.warns(DeprecationWarning) as warninfo:
+            s.remove_background_from_file(background=backgrounds[0][0])
+        assert 'depreciated' in warninfo[0].message.args[0][:]
```
jlaehne, line 79: as above

lumispy/signals/luminescence_spectrum.py
```diff
@@ -512,8 +512,8 @@ def remove_background_from_file(self, background=None, inplace=False, **kwargs):
         -----
         This function does not work with non-uniform axes.
         """
-        warnings.warn('The use of `remove_background_from_file` is depreciated. '
-                      'Please use `remove_background_signal` from Signal1D object', DeprecationWarning, stacklevel=2)
+        warnings.warn('The use of `remove_background_from_file` is depreciated and will be removed in LumiSpy 1.0. '
```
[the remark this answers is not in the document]
jordiferrero: Haha :/ I always used this word wrong!

lumispy/tests/signals/test_luminescence_spectrum.py
```diff
@@ -71,3 +71,9 @@ def test_warnings(self):
         with pytest.warns(SyntaxWarning) as warninfo:
             s.remove_background_from_file(background=None, display=False)
         assert warninfo[0].message.args[0][:18] == "Using the Hyperspy"
+
+    def test_depreciation_warning(self):
```
jlaehne, line 75: `deprecation` again

jlaehne:
> @jlaehne Not it is ready for review. Can you maybe help me tell how to fix the "linting" error? That's something new to me :-)

1. It is an issue with the `black` formatting. In this case probably too long lines and easy to fix. I usually run `black xy.py` on the failing files to fix it automatically.

lumispy/tests/signals/test_luminescence_spectrum.py
```diff
@@ -68,6 +68,10 @@ def test_errors_raise(self):
     def test_warnings(self):
         pytest.importorskip("hyperspy_gui_ipywidgets")
         s = LumiSpectrum(np.ones(50))
-        with pytest.warns(SyntaxWarning) as warninfo:
+        with pytest.warns(SyntaxWarning, match='Using the Hyperspy') as warninfo:
```
jlaehne, line 71: no need for the `as warninfo` (here and below) then, because we do not need to check that variable

jlaehne:
You can also try to set up an automatic black formatting on your system. See also:
https://hyperspy.org/hyperspy-doc/current/dev_guide/coding_style.html
https://kikuchipy.org/en/stable/contributing.html#code-style

PR #116, opened 2022-03-22
Update metadata used in scale_by_exposure

jlaehne:
### Description of the change
- update `scale_by_exposure` according to #109
- some tests can only run with Hyperspy v1.7 (they work locally with the development branch)
- added extra comments in contributors guide

### Progress of the PR
- [x] added tests (partly still commented out),
- [x] activate remaining tests once HyperSpy v1.7 is out,
- [x] ready for review.

### Note
Tests run locally, should work once HyperSpy v1.7 is released.  Merge will need to wait for the release, but can be reviewed already for a speedy release of LumiSpy v0.2 after the HyperSpy release.

### DIFF
--- a/lumispy/signals/common_luminescence.py
+++ b/lumispy/signals/common_luminescence.py
@@ -96,9 +96,10 @@ class CommonLumi:
         Parameters
         ----------
         exposure : float
-            Exposure time in s. If not given, the function tries to find
-            'exposure' or 'dwell_time' in the metadata (for the moment only at
-            Gatan specific nodes).
+            Exposure time in s. If not given, the function tries to use the
+            'metadata.Acqusition_instrument.Detector.integration_time' field or
+            alternatively find any 'integration_time', 'exposure' or
+            'dwell_time' fields in the metadata.
         inplace : boolean
             If `False` (default), a new signal object is created and returned.
             If `True`, the operation is performed on the existing signal object.
@@ -118,18 +119,26 @@ class CommonLumi:
 
         # Make sure exposure is given or contained in metadata
         if isnan(exposure):
-            # use nested_get from hyperspy when it is available
-            if self.metadata.has_item("Acquisition_instrument.CL.exposure"):
+            if self.metadata.has_item(
+                "Acquisition_instrument.Detector.integration_time"
+            ):
                 exposure = float(
-                    self.metadata.get_item("Acquisition_instrument.CL.exposure")
+                    self.metadata.get_item(
+                        "Acquisition_instrument.Detector.integration_time"
+                    )
                 )
-            elif self.metadata.has_item("Acquisition_instrument.CL.dwell_time"):
+            # following will work only from hyperspy v1.7
+            elif self.metadata.has_item("integration_time", full_path=False):
                 exposure = float(
-                    self.metadata.get_item("Acquisition_instrument.CL.dwell_time")
+                    self.metadata.get_item("integration_time", full_path=False)
                 )
+            elif self.metadata.has_item("exposure", full_path=False):
+                exposure = float(self.metadata.get_item("exposure", full_path=False))
+            elif self.metadata.has_item("dwell_time", full_path=False):
+                exposure = float(self.metadata.get_item("dwell_time", full_path=False))
             else:
                 raise AttributeError(
-                    "Exposure not given and can not be "
+                    "Exposure (integration time) not given and can not be "
                     "extracted automatically from metadata."
                 )
         if inplace:
--- a/lumispy/signals/common_luminescence.py
+++ b/lumispy/signals/common_luminescence.py
@@ -90,16 +90,17 @@ class CommonLumi:
         if not inplace:
             return s
 
-    def scale_by_exposure(self, exposure=float("nan"), inplace=False):
-        """Scale data in spectrum by exposure (e.g. convert counts to counts/s).
+    def scale_by_exposure(self, integration_time=None, inplace=False, **kwargs):
+        """Scale data in spectrum by integration time / exposure,
+        (e.g. convert counts to counts/s).
 
         Parameters
         ----------
-        exposure : float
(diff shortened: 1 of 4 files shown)


lumispy/signals/common_luminescence.py
```diff
@@ -118,18 +119,26 @@ def scale_by_exposure(self, exposure=float("nan"), inplace=False):
...
+            # following will work only from hyperspy v1.7
+            elif self.metadata.has_item("integration_time", full_path=False):
                 exposure = float(
-                    self.metadata.get_item("Acquisition_instrument.CL.dwell_time")
+                    self.metadata.get_item("integration_time", full_path=False)
                 )
+            elif self.metadata.has_item("exposure", full_path=False):
+                exposure = float(self.metadata.get_item("exposure", full_path=False))
+            elif self.metadata.has_item("dwell_time", full_path=False):
+                exposure = float(self.metadata.get_item("dwell_time", full_path=False))
```
ericpre, line 138: Would it more simple and readable to do something along the line of:
```python
if exposure is None:
   # exposure is not provided, try to get it from metadata
    exposure = self.metadata.get_item("Acquisition_instrument.Detector.integration_time")
if exposure is None:
    # integration_time is not in the metadata, raise an error
    raise ValueError("Provide `exposure` argument or set `integration_time` in the metadata.")
```

Why should other cases (integration time saved elsewhere in the metadata) be supported? The point of defining a metadata specification is avoid this kind of thing!

lumispy/signals/common_luminescence.py
```diff
@@ -96,9 +96,10 @@ def scale_by_exposure(self, exposure=float("nan"), inplace=False):
         Parameters
         ----------
         exposure : float
-            Exposure time in s. If not given, the function tries to find
-            'exposure' or 'dwell_time' in the metadata (for the moment only at
-            Gatan specific nodes).
+            Exposure time in s. If not given, the function tries to use the
```
ericpre, line 99: Rename the `exposure` argument to `integration_time` so that it is consistent with the metadata specification?

jlaehne: Yes, it would make sense, but I would not want to rename the function not to break the API and would probably allow `exposure` as backwards compatible Alias.

lumispy/signals/common_luminescence.py
jlaehne: > Why should other cases (integration time saved elsewhere in the metadata) be supported? The point of defining a metadata specification is avoid this kind of thing!

Because, even though we have a metadata convention, not all readers parse the metadata yet. But in fact, we should then search the `original_metadata` and not the `metadata`.

lumispy/signals/common_luminescence.py
jlaehne: Added deprecation warning.

lumispy/signals/common_luminescence.py
```diff
@@ -116,27 +122,45 @@ def scale_by_exposure(self, exposure=float("nan"), inplace=False):
...
-            if self.metadata.has_item("Acquisition_instrument.CL.exposure"):
-                exposure = float(
-                    self.metadata.get_item("Acquisition_instrument.CL.exposure")
+        # Make sure integration_time is given or contained in metadata
+        if integration_time is None:
+            if "exposure" in kwargs:
+                integration_time = kwargs["exposure"]
+                raise DeprecationWarning(
+                    "The `exposure` argument will be "
+                    "removed in LumiSpy 1.0, use `integration_time` instead."
```
ericpre, line 131: Same as above: better to use `rename`?

lumispy/signals/common_luminescence.py
ericpre: The readers need to be fixed, then! :)
The motivation for the metadata specification is for user *and* hyperspy functionalities to know where to find relevant information and avoid using heuristic to find it which in my opinion is a bad practice and it needs to be avoided. I don't think, that this is explained in the hyperspy user guide, which would be good to do.

Considering that lumispy doesn't have a stable API at the moment, I would not worry too much about rough edges like this, where the expose needs to be passed explicitly and favour the use of good practice.

lumispy/signals/common_luminescence.py
```diff
@@ -107,6 +109,10 @@ def scale_by_exposure(self, exposure=float("nan"), inplace=False):
         -----
         Sets `metadata.Signal.scaled` to `True`. If intensity units is 'counts',
         replaces them by 'counts/s'.
+
+        .. deprecated:: 0.2
+          The `exposure` argument was renamed `integration_time`, and its use
+          will be deprecated in LumiSpy 1.0.
```
ericpre, line 115:
```suggestion
          The `exposure` argument was renamed `integration_time`, and it will
          be removed in LumiSpy 1.0.
```
The deprecation starts now, as we discourage to use it but it is still possible to use until it is removed in 1.0.

lumispy/signals/common_luminescence.py
```diff
@@ -116,27 +122,32 @@ def scale_by_exposure(self, exposure=float("nan"), inplace=False):
...
-            if self.metadata.has_item("Acquisition_instrument.CL.exposure"):
-                exposure = float(
-                    self.metadata.get_item("Acquisition_instrument.CL.exposure")
+        # Make sure integration_time is given or contained in metadata
+        if integration_time is None:
+            if "exposure" in kwargs:
+                integration_time = kwargs["exposure"]
+                raise DeprecationWarning(
+                    "The `exposure` argument was renamed `integration_time`"
+                    "and its use will be deprecated in LumiSpy 1.0."
```
ericpre, line 131:
```suggestion
                    "The `exposure` argument was renamed to `integration_time`"
                    "and its use will be removed in LumiSpy 1.0."
```
Same as above.

ericpre (approved):
Looks good, I suggested a clearer wording onthe deprecation to avoid confusing about when the deprecation starts and ends and when the `exposure` will stop working.

Issue #5, opened 2019-11-19
Class hierarchy


dnjohnstone:
An issue to discuss openly class hierarchy for lumispy, from an e-mail with @jlaehne 

We want to start out with CL, but need a more general parent that allows to add PL and EL at some point. Also you probably want to add TR-CL/PL at some point.

In general, I guess something like

-lumispec (inheriting from hyperspy/signal1D)
  |-cl
    |-cl-sem
    |-cl-stem
  |-pl
  |-el
-trlumi (inheriting from hyperspy/signal2D ?)
  |-trcl
  |-trpl

jlaehne:
Well cl-sem and cl-stem would inherit from cl, and sticking to the naming of related hyperspy signals, that would give e.g. (with aliases in [ ]):

-LumiSpectrum (inheriting from hyperspy/signal1D) [LuminescenceSpectrum]
|-CLSpectrum          [CL,cathodoluminescence]
||-CLSEMSpectrum   [CLSEM]
||-CLSTEMSpectrum [CLSTEM]
|-PLSpectrum          [PL]
|-ELSpectrum          [EL]
-LumiTransient (inheriting from hyperspy/signal2D ?) [or TRLumi]
|-CLTransient    [TRCL]
|-PLTransient    [TRPL]

`Transient` could also be replaced by `TimeRes`. What would you prefer?

We cans till define several aliases, such

jordiferrero:
I will work on the hierarchy and I personally like the name `TimeRes` more than `Transient`.

- [ ] Modify class hierarchy (by @jordiferrero)

jlaehne:
I thought that `CLTransient` is more similar to `CLSpectrum`, with `TR...` and possibly also `TimeResolved...` as aliases.

dnjohnstone:
I'm a little confused about the time resolved stuff and dimensionality... That's probably because I don't do it!

Why is time-resolved stuff not <t|E> i.e. a Signal1D with a time navigation axis?

Do you do time and spatially resolved stuff? E.g. <x,y|t,E>?

To move forward in a sensible way. I think we'll establish the code hierarchy for all the non-time-resolved stuff now, and make the 0.0.1 release and then we can add the time resolved stuff when ready.

jlaehne:
Well that is why I put it to discussion. Is the time an additional navigation or signal axis? In principle, I would say it could be both and we have to find the most suitable convention.

[...]

PR #118, opened 2022-04-09
reorganize transient classes

### LINKED ISSUE #115: Transient signal classes
We missed something when introducing the transient signal classes last year (and no one seems to have used them so far? Anyway there are no specific functions). At the moment, we have a set of Signal2D classes for transients:
```
LumiTransient | 2 | Luminescence | real | TRLumi, TR luminescence, time-resolved luminescence
CLTransient | 2 | TRCL | real | TR cathodoluminescence, time-resolved cathodoluminescence
PLTransient | 2 | TRPL | real | TR photoluminescence, time-resolved photoluminescence
```

That is correct for a streak camera image <x,y|E,t>, but not for a pure transient <x,y|t> (without spectral information).

The `...Transient` classes should in fact be Signal1D, while there should be an additional set of classes `...TransientSpectrum` for Streak camera images. Any oppinions? I would propose to fix it for the v0.2 release.

(see original discussion in https://github.com/LumiSpy/lumispy/issues/5#issuecomment-558125719 )

jlaehne:
### Description of the change
Closes #115

Split Transient classes into 1D Transient and 2D TransientSpectrum classes. Add `CommonTransient` class for functionality that will work on the time axis.

In a separate PR, functionalities of `LuminescenceSpectrum` class that are important also for streak camera data, will need to be moved to `CommonLuminescence`.

### Progress of the PR
- [X] docstring updated (if appropriate),
- [X] added tests,
- [X] added line to CHANGELOG.md,
- [X] ready for review.


jordiferrero:
Thanks Jonas! Indeed, nobody has really used these classes quite yet (I am working on some functions for the PL transient 1D.

My only comment, following what @LMSC-NTappy  said, is that I would rather keep the names close to the signal features and not the detector. So what about calling it `cl_transient_1d` or `cl_transient_2d` instead of `transient` and `streak`?
Maybe I am getting confused here, but streak implies a streak camera, right? Is there a way of getting 2d luminescence data with other detectors? Idk

jlaehne:
Well, these are 'just' the file names and not the class names, which are e.g. `TransientSpec`, but I could do the same and use `transientspec` in the file names. 

I'm not aware of any other method to obtain spectrally resolved time traces.

LMSC-NTappy (commented):
Looks good to me!

Regarding time-resolved luminescence, I think it can be done using TCSPC and serial spectral acquisition, gated detectors or regular detectors and pump-probe experiment schemes. Not sure what is the dominant 

Only suggetion: I think the file in the signals/ directory could be changed to reflect the new class names.

Cheers

Nicolas

jlaehne:
> Only suggetion: I think the file in the signals/ directory could be changed to reflect the new class names.

Done in https://github.com/LumiSpy/lumispy/pull/118/commits/711b236fb1007a83841fe08079315fe60bb0319c

ericpre (commented):
I went back to the original discussion and add a comments in https://github.com/LumiSpy/lumispy/issues/5#issuecomment-1107456099.
Based on this comment, I would simply suggest to remove the transient classes, for the following reasons:
- It is not clear, what they will be used for
- we should consider them when the need arise
- these `signal_type` can be added later without affecting the existing structure. Actually, already defining a structure could be annoying in the future, in case they needs to be changed!

ericpre (approved):

jordiferrero:
Thanks Eric.
Just to put this into context, we (I recall Duncan and me) created all these classes at the very beginning of LumiSpy to make sure we had a structured "vision" of what the package should look like. I agree no one has had time to dive deeper in the transient classes yet. So I am happy to clean up the package as I think the vision is clear :-)

Issue #133, opened 2022-06-19
Increase coverage


jlaehne:
We have some low hanging fruit in increasing the test coverage of LumiSpy:

- [x] `px_to_nm_grating_solver` in `signals/luminescence_spectrum.py` is not covered [See codecov, lines 291-310]
- [x] For `to_eV` and `to_invcm` in `signals/luminescence_spectrum.py`, the case where the variance is a single `int/float` is not covered when `inplace=True`. [See codecov, lines 154-161 & 309-316](https://app.codecov.io/gh/LumiSpy/lumispy/blob/7eed9e4e96e2e24379ceadb34a8349938b101338/lumispy/signals/luminescence_spectrum.py)
- [x] `scale_by_exposure` in `signals/common_luminescence.py` is missing coverage for the situation where `("Signal.quantity") == "Intensity (Counts)"` [See codecov, lines 154-155](https://app.codecov.io/gh/LumiSpy/lumispy/blob/7eed9e4e96e2e24379ceadb34a8349938b101338/lumispy/signals/common_luminescence.py)
- [x] `remove_spikes` in `signals/cl_spectrum.py` is missing coverage for `show_diagnosis_histogram` [See codecov, line 92](https://app.codecov.io/gh/LumiSpy/lumispy/blob/7eed9e4e96e2e24379ceadb34a8349938b101338/lumispy/signals/cl_spectrum.py)
- [x] `solve_grating_equation` in `utils/axes.py` is missing coverage for axis being non-uniform `DataAxis` [See codecov, line 485](https://app.codecov.io/gh/LumiSpy/lumispy/blob/7eed9e4e96e2e24379ceadb34a8349938b101338/lumispy/utils/axes.py)

- `remove_background_from_file` misses some tests for warnings/errors, but as it will be deprecated is not worth implementing.

jordiferrero:
Coverage 100%

PR #134, opened 2022-06-19
Improve coverage, fix issue with variance conversion

jlaehne:
### Description of the change
- Fix some of the missing coverage mentioned in #133.
- Fix issue with `to_eV` and `to_invcm` changing the variance of the original signal when the variance is a single `int/float` and `inplace=False` (turned up as `inplace=True` was not covered by the test despite it should have been).
- Remove remnants of old HyperSpy axes scheme in `solve_grating_equation` that turned up as a missing coverage for the line checking the axes type version.

### Progress of the PR
- [x] Change implemented,
- [x] ready for review.

### DIFF
--- a/lumispy/tests/signals/test_cl_spectrum.py
+++ b/lumispy/tests/signals/test_cl_spectrum.py
@@ -69,6 +69,17 @@ class TestCLSpectrum:
         np.testing.assert_almost_equal(s1.data[1, 0, 1], 1, decimal=5)
         np.testing.assert_almost_equal(s1.data[0, 2, 29], 1, decimal=5)
 
+        s3 = s.remove_spikes(show_diagnosis_histogram=True)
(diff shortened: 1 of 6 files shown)


ericpre (approved):

lumispy/tests/signals/test_cl_spectrum.py
```diff
@@ -69,6 +69,17 @@ def test_remove_spikes(self):
         np.testing.assert_almost_equal(s1.data[1, 0, 1], 1, decimal=5)
         np.testing.assert_almost_equal(s1.data[0, 2, 29], 1, decimal=5)
 
+        s3 = s.remove_spikes(show_diagnosis_histogram=True)
```
jordiferrero, line 72: Thanks for implementing this!

PR #138, opened 2022-07-23
Update python version in readthedocs.yml

jlaehne:
Update python version in readthedocs.yml


jlaehne:
Readthedocs build and comment seems to work only for `python 3.8`.

PR #140, opened 2022-09-06
Fix bugs in to_raman_shift

jlaehne:
### Description of the change
The conversion to Raman shift (relative wavenumber) is not working as it should:
- data was inverted compared with axis vector
- label axis as `Raman Shift`
- make `Jacobian=False` default
- inplace=False was working with original data instead of copy

### Progress of the PR
- [x] added tests,
- [x] ready for review.


jordiferrero:
Is this ready for review?

jlaehne:
> Is this ready for review?

I still have to fix the tests. Also currently, it runs the `to_invcm` function, which inverts the data order and then inverts it back ... I want to see if it does not make sense to change that. But as it is a serious bug, I should probably finish the fix first and a smoother solution could come later.

lumispy/signals/luminescence_spectrum.py
```diff
@@ -449,26 +458,34 @@ def to_invcm_relative(self, laser=None, inplace=True, jacobian=True):
...
+                self.metadata.Signal.Noise_properties.variance.data = (
+                    self.metadata.Signal.Noise_properties.variance.isig[::-1].data
+                )
         # create and return new signal
         else:
             s2 = self.to_invcm(inplace=inplace, jacobian=jacobian)
             s2.axes_manager.set_axis(
                 invcmaxis, self.axes_manager.signal_axes[0].index_in_axes_manager
             )
+            s2.data = s2.isig[::-1].data
```
jordiferrero, line 478: Is there a way to not have to repeat the same lines of code on both sides of the if statement (code not DRY)? What if you set your copy as your signal in this case of inplace=True?

lumispy/tests/utils/test_axis-conversion.py
```diff
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+# -*- co`ding: utf-8 -*-
```
jordiferrero, line 1: ?

lumispy/signals/luminescence_spectrum.py
jlaehne: Probably an if/else statement could decide whether to operate on a copy of the signal or a pointer to self ... I would however leave that consolidation to a separate PR. I will open an issue.

jordiferrero (approved):

PR #141, opened 2022-09-18
Update installation instructions

jordiferrero:
### Description of the change
Trying to install lumispy for someone else the other day, I encountered some issues that were not clear from the guide. Have added some clarifications.

### Progress of the PR
- [ ] ready for review.

### DIFF
--- a/doc/source/user_guide/installation.rst
+++ b/doc/source/user_guide/installation.rst
@@ -35,6 +35,9 @@ Now activate the LumiSpy environment and install the package using pip:
 
 Installation is completed! To start using it, check the next section.
 
+.. Note::
+
+    If the installation fails using ``pip``, try installing using ``conda install lumispy -c conda-forge``.
 
 Getting Started
 ===============
@@ -47,4 +50,5 @@ entered into an anaconda prompt (from scratch):
 .. code-block:: bash
 
     (base) conda activate lumispy
+    (lumispy) pip install jupyterlab
     (lumispy) jupyter lab


jlaehne:
We still have the duplicate installation instructions in the user guide and in https://github.com/LumiSpy/lumispy/blob/main/INSTALLATION.md that should be merged together in the page of the user guide.

doc/source/user_guide/installation.rst
```diff
@@ -47,4 +50,5 @@ entered into an anaconda prompt (from scratch):
 .. code-block:: bash
 
     (base) conda activate lumispy
+    (lumispy) pip install jupyterlab
```
ericpre, line 53: It can cause issues to mix conda and pip as pip doesn't ensure ABI and it can break dependencies easily. Why recommending to use pip and not conda?

doc/source/user_guide/installation.rst
```diff
@@ -35,6 +35,9 @@ Now activate the LumiSpy environment and install the package using pip:
 
 Installation is completed! To start using it, check the next section.
 
+.. Note::
+
+    If the installation fails using ``pip``, try installing using ``conda install lumispy -c conda-forge``.
```
ericpre, line 40: I can see two situations where it can happen:
- some lumipsy dependencies can't be installed, for example, compiler missing.
- seems to install OK, but pip broke the python installation because of ABI incompatibility, for example, it is not possible to import numpy, scipy or numba anymore.

Would it be worth being more specific and explain a bit more? As it currently, I don't understand what is the logic.

ericpre (commented):
I suspect that these instructions may do more harm than good - see comments.

Issue #129, opened 2022-06-15
extend functionality of crop_edges


jlaehne:
#### Describe the context
`crop_edges` is a wrapper function for `inav` that takes the parameter `crop_px` and removes as many pixels from all sides of a map.

#### Describe the functionality you would like to see.
To make this function more versatile, it would make sense to alternatively allow the following sets of parameters:
- `crop_x_px` and `crop_y_px` to crop different amounts of pixels from the x and y directions
- `crop_left_px`, `crop_right_px`, `crop_top_px`, `crop_bottom_px` to crop different amounts of pixels from all sides (check in `matplotlib` what order would be typically used for such a set of parameters corresponding to the edges).

#### Additional information
@jordiferrero , you originally implemented this function. Would this extension make sense to you? We have a student that could implement it.

jordiferrero:
This makes sense.
I initially implemented this function because I always run this "simple" cropping during pre-processing.
It would be great for a student to do it. Just make sure the original functionality is still working (so when you only pass 1 parameter, it crops all sides equally).

jlaehne:
Of course, one should be able to pass either of the three sets of parameters mentioned above.

jordiferrero:
Glad to hear a student is coming to help you with coding :-)

Issue #73, opened 2021-04-14
Dealing with SEM/HAADF complementary data with their respective luminescence


jordiferrero:
I have recently been wondering on what the best strategy would be to do the following:

All of my CL-SEM data always comes with SE images with it (I suppose for CL-TEM the equivalent would be a HAADF image). These SE images are acquired at the same time as the luminescence. However, they are not integrated with hyperspy at all.
While it is true that you can load the SE image separately (as `s`) and the use this `s` object as `navigation_axis` for the `cl` object, this all needs to be done manually.
What's worse is that when I do some processing on the `cl` object (e.g. crop or subindex some of the navigation axes) the `s` SE image does not link with the `cl` object in anyway.
Ideally, i would like to link these 2 datasets together.

I have an idea to do that, but I wanted too quickly discuss the approach before I get into the coding.
One way would be to keep the SE image array in the metadata and, every time you apply a certain operation that modifies the navigation axis, then those changes are also applied to the SE metadata (e.g. crop pixels). Moreover, I would like to have an option when plotting where we can specify to use the SE image as navigation axis instead of the default "panchromatic".

Can you think of any better ways to implement such feature?

ericpre:
The management of the different signals is taken care by the workflow: the python script or the notebook.

In the case of the navigator, it would be worth considering improving support signal navigator if this is thought in a generic fashion and in a way that it doesn't too much burden to other functions. Maybe it is worth continuing this part of the discussion at https://github.com/hyperspy/hyperspy/issues/2639?

jordiferrero:
So you mean I should stick to keeping track of the changes in the navigation axis of the `cl` signal and manually applying the same changes to the `s` SE image file?
That's what I am doing, but I still think there must be a better way...

jlaehne:
@jordiferrero - I understand your point. In principle it applies also for EELS and EDX data and might be of more general interest. Though in many cases the spectral maps will have a lower resolution than proper HAADF or SE images of the same region, and a direct linking of this data is of limited value.

For example, our CL system does not record the SE signal in parallel. I usually have a corresponding SE image taken beforehand and the spectral image corresponds to only a ROI on that image. That SE image usually has a better resolution than the spectral image as I hardly ever use step sizes below 20 nm.

[...]

PR #142, opened 2022-09-18
Crop edges update

jordiferrero:
### Description of the change
Following the request to extend the functionality of crop_edges from #129. I have implemented the following changes:
- Function is more versatile (it can take 1, 2, or 4 crop values) and can crop in individual directions.
- Function can also crop in percentages wrt the width/height of the navigation axis
- Function can crop multiple signals in the same fashion (e.g. luminescence map and reference SE signal)

### Progress of the PR
- [x] Change implemented,
- [x] docstring updated,
- [x] update user guide (utilities page),
- [x] added tests,
- [x] ready for review.


jordiferrero:
Does lint correct formatting when PR is merged? I have tried to correct it manually but I cannot seem to get it working without an error

jlaehne:
> Does lint correct formatting when PR is merged? I have tried to correct it manually but I cannot seem to get it working without an error...

No, it is only doing the check. But you can run `black` locally to automatically reformat a file accordingly, see
https://hyperspy.org/hyperspy-doc/current/dev_guide/coding_style.html

lumispy/signals/common_luminescence.py
```diff
@@ -21,47 +21,89 @@
...
-        if crop_px * 2 > width or crop_px * 2 > height:
+        units_accepted = ("px", "pixel", "percent", "%")
+        if crop_units.lower() not in units_accepted:
             raise ValueError(
-                "The pixels to be cropped cannot be larger than half the width or the length!"
+                "The param crop_units only accepts `pixel` units or `percent`."
             )
+
+        w = self.axes_manager.shape[0]
+        h = self.axes_manager.shape[1]
```
jlaehne, line 57: for a linescan, this would be the signal axis.

lumispy/signals/common_luminescence.py
```diff
@@ -21,47 +21,89 @@
...
+    def crop_edges(self, crop_range, crop_units="pixel"):
         """Crop the amount of pixels from the four edges of the scanning
-        region, from out the edges inwards.
+        region, from out the edges inwards. Cropping can happen uniformily or by specifying the croping range for each axis or each side.
 
         Parameters
         ----------
-        crop_px : int
-            Amount of pixels to be cropped on each side individually.
+        crop_range : int, float, tuple
```
jlaehne, line 38: renaming breaks the API, we should still accept `crop_px` with a warning that it is deprecated and will be removed in v1.0

ericpre:
Would it be worth fixing the test suite first?

How different is it from doing the following?

```python
# remove one pixel on both side
s.isig[1:-1]

# remove one units on both side
s.isig[1.0:-1.0]

# remove one eV on both side
s.isig["1.0 eV":"-1.0eV"]
```

The last two seem to be broken but would be expected to work. The same approach would work for `inav`. The name `crop_edges` is ambiguous as it is not clear it it refers to navigation or signal space - my first intention was that it was in signal space.

lumispy/signals/common_luminescence.py
```diff
@@ -21,47 +21,89 @@
...
+                f"The crop_range value must be a number or a tuple, not a {crop_range_type}"
+            )
 
-        # Store transformation in metadata (or update the value if already previously transformed)
+        # Negative means reverse indexing
+        crop_vals = array(crop_vals) * [1, -1, -1, 1]
+
+        # Convert percentages to pixel units
+        if crop_units.lower() in units_accepted[-2:]:
+            crop_vals = crop_vals * array([w, h] * 2)
```
jlaehne, line 82: This assumes that a percentage is given as value between 0 and 1, which is not clear from the docstring (especially as the unit could be `%`)!

lumispy/signals/common_luminescence.py
```diff
@@ -21,47 +21,89 @@
...
+            crop_vals = crop_vals.astype(int)
```
jlaehne, line 83: Rounds downwards, which is a sensible choice, but might be mentioned in the docstring?

lumispy/signals/common_luminescence.py
```diff
@@ -21,47 +21,89 @@
...
 
+        # Remove 0 for None
```
jlaehne, line 85: We should maybe also accept `None` as value in the original tuple? Needs to be tested for as currently some of the above lines would throw an error.

lumispy/signals/common_luminescence.py
```diff
@@ -21,47 +21,89 @@
...
+        crop_ids = [x if x != 0 else None for x in crop_vals]
+
+        # Crop accordingly
+        signal_cropped = self.inav[
+            crop_ids[0] : crop_ids[2], crop_ids[3] : crop_ids[1]
+        ]
+
+        # Check if cropping went too far
+        if 0 in signal_cropped.axes_manager.navigation_shape:
+            warn(
```
jlaehne, line 95: Shouldn't that possibly return an error?

lumispy/signals/common_luminescence.py
```diff
@@ -21,47 +21,89 @@
...
+                "The pixels to be cropped surpassed the width/height of the signal navigation axes.",
+                UserWarning,
+            )
+
+        # Store transformation in metadata (or update the value if already previously transformed)
         try:
-            signal_cropped.metadata.Signal.cropped_edges += crop_px
+            signal_cropped.metadata.Signal.cropped_edges += abs(crop_vals)
         except AttributeError:
```
jlaehne, line 103: This shouldn't lead to an exception. You can use `has_item` to check if the metadata entry exists: http://hyperspy.org/hyperspy-doc/current/user_guide/metadata_structure.html#searching-for-keys

lumispy/tests/signals/test_common_luminescence.py
```diff
@@ -17,24 +17,57 @@
 # along with LumiSpy. If not, see <https://www.gnu.org/licenses/#GPL>.
 
 import numpy as np
-import pytest
-
+from pytest import raises, mark, warns
+from warnings import WarningMessage
```
jlaehne, line 21: Is this import needed?

jlaehne:
> Would it be worth fixing the test suite first?

Yes, would make the review easier.

> The last two seem to be broken but would be expected to work. The same approach would work for `inav`. The name `crop_edges` is ambiguous as it is not clear it it refers to navigation or signal space - my first intention was that it was in signal space.

The last two might be broken for non-uniform axes? We should check that.

Indeed, it is a somewhat complicated wrapper for `inav`, which I personally have so far always used for this functionality (and we need to be clear that it is about the navigation axes). Added value of the latest implementation that I see would be that it accepts percentage values additionally and that the cropping is documented in the metadata.

ericpre:
> Added value of the latest implementation that I see would be that it accepts percentage values additionally and that the cropping is documented in the metadata.

I don't know if this is the kind of thing that should be documented in the metadata because I think that this is better to record this information in the script or the notebook itself. The issue with adding this kind of processing information to the metadata (other than the one which we know will be used for further processing, in which case, we can't get away with it) is that it is very brittle - difficult to enforce consistency and to maintain.

jordiferrero:
> > Added value of the latest implementation that I see would be that it accepts percentage values additionally and that the cropping is documented in the metadata.
> 
> I don't know if this is the kind of thing that should be documented in the metadata because I think that this is better to record this information in the script or the notebook itself. The issue with adding this kind of processing information to the metadata (other than the one which we know will be used for further processing, in which case, we can't get away with it) is that it is very brittle - difficult to enforce consistency and to maintain. More generally, this touch on the topic of data management, which is I believe is not the concern of library like lumispy, but rather the one of electronic laboratory notebook (ELN) and there are a lot of solution for this.

The reason I wanted to document the cropping in the metadata is because CL data is always paired with SE data (for the SEM). The SE images are normally a different file and, if I want to plot a CL map next to the SE image, then I need to know what cropping was done exactly on the map. This is a "necessary" preliminary function to accomplish #73.
I can correct all the revisions requested (thanks for looking into it) if that makes sense to you.

I could also implement the cropping by units, which is currently not implemented.

jordiferrero:
> Okay, it makes more sense. What would be the corresponding workflow in a few line of code?

I am not 100% sure yet but something like this:

```
def transform_image_to_s_nav_axis(image, s, plot=True):
    read metadata from s on crop + rebinning
    apply crop + rebin to image
    return image_transform & plot with navigator as image
```

I am not sure if there is a way to store image in the s object and always call it as navigation axis when plotting the signal_1d/2d object s.

ericpre:
> I am not 100% sure yet but something like this:
> 
> ```
> def transform_image_to_s_nav_axis(image, s, plot=True):
>     read metadata from s on crop + rebinning
>     apply crop + rebin to image
>     return image_transform & plot with navigator as image
> ```

Would it make sense to pass the other signal at the same time? Pass a list of signal and all apply the cropping to all of them instead of calling the `crop_edge` and relying on the metadata?

> I am not sure if there is a way to store image in the s object and always call it as navigation axis when plotting the signal_1d/2d object s.

There is lazy signal and it would be good to do it with non-lazy signal.

jordiferrero:
> > I am not 100% sure yet but something like this:
> > ```
> > def transform_image_to_s_nav_axis(image, s, plot=True):
> >     read metadata from s on crop + rebinning
> >     apply crop + rebin to image
> >     return image_transform & plot with navigator as image
> > ```
> 
> Would it make sense to pass the other signal at the same time? Pass a list of signal and all apply the cropping to all of them instead of calling the `crop_edge` and relying on the metadata?
> 
> > I am not sure if there is a way to store image in the s object and always call it as navigation axis when plotting the signal_1d/2d object s.
> 
> There is lazy signal and it would be good to do it with non-lazy signal.

Based on this feedback, I think the `crop_edges` function should be expanded to accepting a list of hyperspy signals. Then it should crop all the signals equally and return them in the same order as input.
This means that all the functionalities that crop_edges has now are good. All I need to implement is the multiple signals input.
Then this "wrapper for inav" makes more sense, in my opinion.
I will implement this and let you review again.
Thanks

jordiferrero:
> > > I am not 100% sure yet but something like this:
> > > ```
> > > def transform_image_to_s_nav_axis(image, s, plot=True):
> > >     read metadata from s on crop + rebinning
> > >     apply crop + rebin to image
> > >     return image_transform & plot with navigator as image
> > > ```
> > 
> > 
> > Would it make sense to pass the other signal at the same time? Pass a list of signal and all apply the cropping to all of them instead of calling the `crop_edge` and relying on the metadata?
> > > I am not sure if there is a way to store image in the s object and always call it as navigation axis when plotting the signal_1d/2d object s.
> > 
> > 
> > There is lazy signal and it would be good to do it with non-lazy signal.
> 
> Based on this feedback, I think the `crop_edges` function should be expanded to accepting a list of hyperspy signals. Then it should crop all the signals equally and return them in the same order as input. This means that all the functionalities that crop_edges has now are good. All I need to implement is the multiple signals input. Then this "wrapper for inav" makes more sense, in my opinion. I will implement this and let you review again. Thanks

This is now ready for a final review. All tests and coverage had been sorted out @jlaehne @ericpre

doc/source/user_guide/utilities.rst
```diff
@@ -81,17 +81,27 @@ The default operational mode is ``inplace=False`` (a new signal object is return
...
+    >>> signals
+    [CLSpectrum <256,256|1024>, Signal2D <128,128|1>]
+    >>> signals_cropped = lum.utils.axes.crop_edges(crop_range=5, crop_units="%", rebin_nav=True)
+    >>> signals_cropped
+    [CLSpectrum <243,243|1024>, Signal2D <243,243|1>]
 
-*[TODO: add possibility to crop different amounts of pixels on different sides]*
+.. Note::
 
+    Many scanning luminescence techniques result in edge defects at the edges of the scanned region. This funciton enables the same cropping of the navigation axis for a list of signals in the same region to correct for such defect.
```
jlaehne, line 100:
```suggestion
    Many scanning luminescence techniques result in edge defects at the edges of the scanned region.
    This function enables the same cropping of the navigation axis for a list of signals in the same region
    to correct for such defect.
```

jlaehne: line wrapping applies also to other parts of the documentation.

lumispy/tests/utils/test_axes.py
```diff
@@ -172,3 +173,171 @@ def test_joinspectra_FunctionalDA(average, scale, kind):
...
+
+
+@mark.parametrize("data", [((10,) * 4), ((10,) * 5)])
+def test_crop_edges_multidim(data):
+    s1 = [LumiSpectrum(ones((data)))]
+    with raises(NotImplementedError, match="navigation axes with more than 2"):
+        crop_edges(s1, 2)
+
+
+def test_crop_edges_depreciated():
```
jlaehne, line 275:
```suggestion
def test_crop_edges_deprecated():
```

lumispy/utils/axes.py
```diff
@@ -176,6 +179,179 @@ def var2invcm(variance, factor, invcmaxis, ax0=None):
...
+# navigation axis manipulation
+#
+
+
+def crop_edges(
+    S,
+    crop_range=None,
+    crop_units="pixels",
+    rebin_nav=False,
+    *,
```
jlaehne, line 192: Is the star a kind o `kwargs` ? Shouldn't it come last, if at all?

lumispy/utils/axes.py
```diff
@@ -176,6 +179,179 @@ def var2invcm(variance, factor, invcmaxis, ax0=None):
...
+    crop_units : str
+        Select in which units cropping happens. Value can be either ``pixels``/``px`` (default),
+        or ``percent``/``%``. All units are rounded downwards.
+
+    rebin_nav : bool
+        If the navigation axes shape is different between signals in the list S, all signals will be rebinned to match the shape of the first signal in the list.
+
+    Returns
+    -------
+    S_cropped : list of the respective cropped HyperSpy Signal objects or a cropped single Signal if only one signal object is passed as input.
```
jlaehne, line 222: type should be more concise like `Signal or list of Signals`
The details can come in the descriptive line after.

lumispy/utils/axes.py
```diff
@@ -176,6 +179,179 @@ def var2invcm(variance, factor, invcmaxis, ax0=None):
...
+        A list of smaller, cropped Signal objects.
+    """
+    # Depreciation warning (for compatibility with ``crop_px``)
+    if crop_range is not None and crop_px is not None:
+        raise TypeError(
+            "Both ``crop_range`` and the depreciated ``crop_px`` were passed. Use only ``crop_range``."
```
jlaehne, line 228:
```suggestion
            "Both ``crop_range`` and the deprecated ``crop_px`` were passed. Use only ``crop_range``."
```

lumispy/utils/axes.py
```diff
@@ -176,6 +179,179 @@ def var2invcm(variance, factor, invcmaxis, ax0=None):
...
+        if signal_cropped.metadata.has_item("Signal.cropped_edges"):
+            signal_cropped.metadata.Signal.cropped_edges += abs(crop_vals_s)
+        else:
+            signal_cropped.metadata.set_item("Signal.cropped_edges", abs(crop_vals_s))
+        S_cropped.append(signal_cropped)
+
+    if no_list:
+        S_cropped = S_cropped[0]
+
+    return S_cropped
```
jlaehne, line 352:
```suggestion
    if no_list:
        return S_cropped[0]
    else:
        return S_cropped
```

lumispy/utils/axes.py
```diff
@@ -176,6 +179,179 @@ def var2invcm(variance, factor, invcmaxis, ax0=None):
...
+    rebin_nav : bool
+        If the navigation axes shape is different between signals in the list S, all signals will be rebinned to match the shape of the first signal in the list.
+
+    Returns
+    -------
+    S_cropped : list of the respective cropped HyperSpy Signal objects or a cropped single Signal if only one signal object is passed as input.
+        A list of smaller, cropped Signal objects.
+    """
+    # Depreciation warning (for compatibility with ``crop_px``)
+    if crop_range is not None and crop_px is not None:
```
jlaehne, line 226: Could raise a warning instead of an error and just use `crop_range`?

lumispy/utils/axes.py
```diff
@@ -176,6 +179,179 @@ def var2invcm(variance, factor, invcmaxis, ax0=None):
...
+#
+
+
+def crop_edges(
+    S,
+    crop_range=None,
+    crop_units="pixels",
+    rebin_nav=False,
+    *,
+    crop_px=None,
```
jlaehne, line 193: It is a bit weird to have a deprecated keyword in the header!

lumispy/utils/axes.py
```diff
@@ -176,6 +179,179 @@ def var2invcm(variance, factor, invcmaxis, ax0=None):
...
+        If a number or a tuple of size 1 is passed, all sides are cropped by the
+        same amount. If a tuple of size 2 is passed (``crop_x``, ``crop_y``), a different
+        amount of pixels/percentage is cropped from the x and y directions,
+        respectively. If a tuple of size 4 is passed (``crop_left``, ``crop_bottom``,
+        ``crop_right``, ``crop_top``), a different amount of pixels/percentage is
+        cropped from each edge individually.
+
+    crop_units : str
+        Select in which units cropping happens. Value can be either ``pixels``/``px`` (default),
+        or ``percent``/``%``. All units are rounded downwards.
```
jlaehne, line 215: a `unit` cannot be rounded, just the `value`

jlaehne:
The operation on a list is definitely an added value! Did some basic trials on some of my data and it worked nicely.

lumispy/utils/axes.py
ericpre, line 215: The API could be simplify by merging the `crop_ranges` and `crop_units` arguments, similarly as this is done with the [`inav` and `isig`](https://hyperspy.readthedocs.io/en/stable/user_guide/signal.html#indexing):
- use `int` for indices (pixel)
- use `float` to use the axis values.
- use string starting with `rel`, e.g. `rel0.1` to use relative coordinate, i.e. the equivalent of percent
- use string finishing with an unit to to unit conversion on the fly.

This is in implemented in the axis, so this should be easy to re-use it.

doc/source/user_guide/utilities.rst
```diff
@@ -81,17 +81,27 @@ The default operational mode is ``inplace=False`` (a new signal object is return
...
 It is a convenience wrapper for the ``inav`` `method in HyperSpy
 <https://hyperspy.org/hyperspy-doc/current/user_guide/signal.html#indexing>`_.
 
 .. code-block:: python
 
-    >>> s.crop_edges(crop_px=2)
+    >>> signals = [cl_map, sem_image]
+    >>> signals
+    [CLSpectrum <256,256|1024>, Signal2D <128,128|1>]
+    >>> signals_cropped = lum.utils.axes.crop_edges(crop_range=5, crop_units="%", rebin_nav=True)
```
ericpre, line 94:
```suggestion
    >>> signals_cropped = lum.utils.axes.crop_edges(signals, crop_range=5, crop_units="%", rebin_nav=True)
```

As this is acting on signals and not axes, would it make more sense to have this function define in `lum.utils.crop_edges`?

jlaehne: Yes, also makes it less complicated to call.

lumispy/utils/axes.py
jlaehne: See [crop_signal1D](https://hyperspy.readthedocs.io/en/latest/api/hyperspy._signals.signal1d.html#hyperspy._signals.signal1d.Signal1D.crop_signal1D) on how this is implemented in a similar function (just signal instead of navigation space) in HyperSpy.

An interactive option could be interesting as well.

jordiferrero: Thanks for the comments. 
The `crop_signal1D` and the respective 2d equivalent `crop_image` do not accept sting fancy indexing (e.g. `rel0.1` nor `1nm`). These functions only work with floats or integers.
I can implement them, but the fancy string formatting would not work (which I think it's ok).

Also, I'd like to point out that this function here does a few more checks than the basic signal operators (such as checking for axes shapes, guessing what the cropping ranges should be, and rebins...).

doc/source/user_guide/utilities.rst
jordiferrero: Implemented. Had to shift some of the documentation accordingly. Also, following the same reasoning, I moved the `join_spectra` function to utils too, as it acts on signals and not axes.

doc/source/user_guide/signal_tools.rst
```diff
@@ -143,3 +143,29 @@ The default operational mode is ``inplace=False`` (a new signal object is return
 
     >>> s.remove_negative(0.1)
 
+
```
jlaehne, line 146: These functions do not operate on the signal, so I would like to keep them in the utils page of the documentation.

lumispy/tests/utils/test_axes.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2019-2023 The LumiSpy developers
+# Copyright 2019-2022 The LumiSpy developers
```
jlaehne, line 2:
```suggestion
# Copyright 2019-2023 The LumiSpy developers
```

doc/source/user_guide/signal_tools.rst
jlaehne: (unit conversion and grating equation)

lumispy/utils/__init__.py
```diff
@@ -27,6 +27,386 @@
     axis2invcm,
     data2invcm,
     var2invcm,
-    join_spectra,
     solve_grating_equation,
 )
+
```
jlaehne, line 32: The utilities should not be in the `__init__.py` file! Create a file `signal.py` instead?

jlaehne:
Thanks @jordiferrero for bringing up the issue that some of the filenames and function placements are maybe not so logical.
However, moving functions and related tests other than `crop_edges` should normally be done in a separate PR. Makes it quite complicated to review this PR now.

jordiferrero:
> Thanks @jordiferrero for bringing up the issue that some of the filenames and function placements are maybe not so logical. However, moving functions and related tests other than `crop_edges` should normally be done in a separate PR. Makes it quite complicated to review this PR now.

@jlaehne You are right. This PR has got out of hand... it's OK.
I will open a new PR on this feature (and only this feature) and will close this one in due time. For now, please don't bother reviewing it. Sorry :-)

jlaehne:
I think you could also consolidate this one and move the stuff concerning other functions to a separate PR. But whatever works better for you :-)

Issue #139, opened 2022-08-29
Adding an interactive way to slice HS over a wavelength range, and view the result spatially!


0Hughman0:
#### Describe the functionality you would like to see.

In the context of taking hyperspectral maps, sometimes you may have a sample where in different regions one would expect peaks at different wavelengths to become dominant. 

It's useful to be able to quickly scan over a given wavelength (or energy!) range, and see how the intensity of signals within that range vary spatially.

#### Describe the context

If implemented correctly, this could be used for a number of different hyperspectral signals. 

The Dorset software for Chronos CL systems has this feature, but as far as I know this doesn't exist in Lumispy/ Hyperspy.

#### Proposed Solution

Perhaps this specific sort of analysis is outside of the scope of Lumispy. However, I thought something like this might fit well into the `utils` section of the library.

I have put together a provisional implementation here:

https://github.com/0Hughman0/lumispy/blob/0fc6fa18e82069af116acdfdc066fb244da05387/lumispy/utils/analysis.py#L4

I've called the feature 'AutoPeakMap'... not sure if this is the best name!

An example would be:

```
from lumispy.utils.analysis import AutoPeakMap, RangeChannel, plot_auto_peakmap

cl = hs.load(path, signal_type='CL_SEM')

apm = AutoPeakMap(cl,
                  RangeChannel('red', range_=[730, 760]),
                  RangeChannel('green', range_=[300, 400]))
apm.plot()

```

Renders two plots, the 'navigator', in the hyperspy sense:

![nav](https://user-images.githubusercontent.com/16800307/187270295-c65d6bb6-143e-47a4-a247-778d9d670aa6.png)

and the 'signal':

![sig](https://user-images.githubusercontent.com/16800307/187270341-7db31791-9ec2-49eb-b5e9-e7dcb52e5bcd.png)

The ranges in the 'navigator' can be moved around interactively and the signal plot will update accordingly.

For more quick and dirty analysis I implemented a convenience method:

```
plot_auto_peakmap(cl)
```

Which chooses sensible (enough) defaults for the `RangeChannel`s.

If this is something you think could slot into LumiSpy, I'll go ahead and write the tests and expand the docstrings etc.

Feedback on the implementation/ naming more than welcome!

Thanks,

Hugh

ericpre:
This makes sense to have such features; as an alternative of implementing new classes, I would suggest to use already existing functionalities using the following approach:

```python
[...]

PR #148, opened 2022-09-21
Implemented plot_span_map function

0Hughman0:
### Description of the change
- Added analysis section to utils folder. I would suggest this could be a place to put utilities for creating interactive plots that allow users to analyse their data.
- implemented plot_span_map. This is a tool for automatically plotting a spatial map of intensity of a hyperspectral map over a number of ranges defined by a SpanROI.

### Progress of the PR
- [x] update user guide (if appropriate),
- [x] added tests,
- [x] ready for review.

```python
import numpy as np
import hyperspy.api as hs

from lumispy.utils.analysis import plot_span_map

s = hs.signals.Signal1D(data=np.random.random((64, 64, 1024)), 
                        axes=[{'name': 'x', 'size': 64}, 
                              {'name': 'y', 'size': 64}, 
                              {'name': 'sig', 'size': 1024}])

plot_span_map(s, nspans=2)
```

Capture


jlaehne:
Sorry, still did not find the time for a detailed look at the code. But already the demo video is very cool.

As suggested by @ericpre in https://github.com/LumiSpy/lumispy/issues/139#issuecomment-1231397069 it would be quite useful for any hyperspectral signal (EDX, EELS) and not only luminescence - so we should at least discuss incorporating it upstream.

For linescans, I could imagine an equivalent tool (possibly covered by the same function) that plots intensity traces for the chosen spectral ROIs.

0Hughman0:
Great thanks. 

With some guidance I would be happy to help adapt the code to be incorporated into hyperspy.

I think knowing how best to handle the various shapes of data will be the challenge here.

lumispy/utils/analysis.py
```diff
@@ -0,0 +1,117 @@
...
+        Sum over all positions of `sig`, the 'navigator' for `plot_span_map`
+    spans: hs.roi.SpanROI
+        The span ROIs from the navigator
+    span_sigs: hs.signals.BaseSignal
+        slices of `sig` according to each span roi
+    span_sums: hs.signals.BaseSignal
+        the summed `span_sigs`.
+    """
+    if nspans > 3:
+        raise ValueError("Maximum number of spans allowed is 3")
```
jordiferrero, line 51: Is there a reason why 3 is the limit?

jordiferrero: If so, it should be stated in the documentation.

lumispy/utils/analysis.py
```diff
@@ -0,0 +1,117 @@
...
+    if (
+        sig.axes_manager.signal_dimension != 1
+        or sig.axes_manager.navigation_dimension != 2
+    ):
+        sig_dims, nav_dims = (
+            sig.axes_manager.signal_dimension,
+            sig.axes_manager.navigation_dimension,
+        )
+        raise ValueError(
+            f"This method is designed for data with 1 signal and 2 navigation dimensions, not {sig_dims} and {nav_dims} respectively"
```
jordiferrero, line 62: I think you could easily also make it work for a 1 signal 1 nav axis with little work.

lumispy/utils/analysis.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+
+    colors = ["red", "green", "blue"]
```
jordiferrero, line 65: Could you have this as an argument too?  Happy to have these 3 as default, but it would be nice to be able to change that.

lumispy/utils/analysis.py
```diff
@@ -0,0 +1,117 @@
...
+    Returns
+    -------
+    all_sum: hs.signals.BaseSignal
+        Sum over all positions of `sig`, the 'navigator' for `plot_span_map`
+    spans: hs.roi.SpanROI
+        The span ROIs from the navigator
+    span_sigs: hs.signals.BaseSignal
+        slices of `sig` according to each span roi
+    span_sums: hs.signals.BaseSignal
+        the summed `span_sigs`.
```
jordiferrero, line 48: Can you also add as argument the key `ax` to select to plot the resulting map to an already existing matplotlib `figure` object? See the plotting hyperspy functions for reference. I think that is very useful, especially when you want to edit all the figure/axis parameters yourself (e.g. imagine I want the imshow plot to be in log scale instead...)

jordiferrero (changes requested):

lumispy/tests/utils/test_plot_span_map.py
```diff
@@ -0,0 +1,158 @@
...
+        ],
+    )
+
+    sig.set_signal_type("CLSEM")
+    return sig
+
+
+def sig_mpl_compare(type: set(["sig", "nav"])):
+    def wrapper(f):
+        @pytest.mark.mpl_image_compare(baseline_dir="plot_span_map")
```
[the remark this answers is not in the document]
0Hughman0: Thanks for the comments Jordi!

I am trying to follow the convention set out by hyperspy i.g. for testing the drawing library they make use of these kinds of tests:

https://github.com/hyperspy/hyperspy/tree/RELEASE_next_minor/hyperspy/tests/drawing/plot_model1d

My concern is with the interactivity, I think it would be good to ensure that the appearance of the plots updates as the spans are moved around, which I think the current tests do.

lumispy/utils/analysis.py
```diff
@@ -0,0 +1,117 @@
...
+#
+# LumiSpy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with LumiSpy. If not, see <https://www.gnu.org/licenses/#GPL>.
+
+import hyperspy.api as hs
```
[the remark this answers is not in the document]
0Hughman0: That sounds fair to me. 

I'm not super familiar with where things live and where is best to put things.

lumispy/utils/analysis.py
0Hughman0: Err not especially. It's just that beyond 3, it's a bit more challenging to come up with a sensible colour scheme. 

Using `'red'`, `'green'` and `'blue'` is helpful because we can map these directly to corresponding cmaps: `Reds`, `Greens` and `Blues`: 

https://matplotlib.org/stable/tutorials/colors/colormaps.html?highlight=reds#sequential

Looks like in theory could add purple, orange and gray.

To allow for arbitrary colour combinations would be possible, but would allow for some smart code to map between span colours and appropriate cmaps.

lumispy/utils/analysis.py
0Hughman0: I can see why this might be useful, and I think it could be added, however I think this is still possible by using the signals returned by the function.

You can perform the slicing using the `plot_span_map` method and then use the returned signal objects to plot things over whichever axes you would like.

I'd argue it could get quite confusing as to which ax is which etc. and we could end up with many arguments.

Not sure what the consensus is, especially if this is the convention.

This has made me realise that we should probably provide a `spans` argument which allows a user to provide a list of `SpanROI`s to use to slice your data, with ready defined ranges. This is the sort of thing I do often - use interactive ROIs to explore and find a desired ROI, and then overwrite the previous ROI definition to make the default position the one I just found... if that makes sense!

lumispy/utils/analysis.py
0Hughman0: I'm trying to get my head around what this would look like.

We have a line of spectra, the 'nav' is the sum over these spectra, then the signal is a line showing the intensity at each point.

tbf this might work already, just have to remove the type check!... and write the tests.

I'd just be a bit worried that then the docstring is would start to get quite confusing - is it still `plot_span_map`, or is that more `plot_span_line`!?

This is maybe an argument for adding this as a method to the various signal types defined in lumispy.

lumispy/utils/analysis.py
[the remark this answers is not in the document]
jlaehne: As most visualizations reside in `plot` in hyperspy, that could indeed make sense.

lumispy/utils/analysis.py
jlaehne: Indeed, that is the linescans I mentioned above. You select regions in the spectrum and then get a lineplot of the intensity against position. There would be a difference in the plotting, I guess, as it is not a map any more.

The signal types in lumispy can all have 0, 1 or 2 navigation dimensions. So a differentiation does not make sense. Only, time-resolved data has in theory an additional signal axis, which complicates things (but in the long term it would be cool to also mark regions along the time trace and get the maps for that).

lumispy/utils/analysis.py
jordiferrero: Could you try to implement that, following the hyperspy plot colourmaps [here](http://hyperspy.org/hyperspy-doc/current/user_guide/visualisation.html#id28). I believe you can set as default n=3 with RGB colours, or if n>3, then apply the `'mpl_colors' camp. Or if the user sets their own, you can convert the continuous cmap to a discrete n colour list (easily done with matplotlib palets functions.

lumispy/utils/analysis.py
jordiferrero: I agree. I would definitely make sure it works for 0, 1 and 2 dimensional navigation axis (0 returns single value, 1 returns a line plot, 2 returns an image).
As for more than 1 dimensional signal axes, I think it goes beyond this PR.

lumispy/utils/analysis.py
jordiferrero: Discussed above.

lumispy/utils/analysis.py
jordiferrero: OK. I understand. The user would need to replot each component independently... For the ax argument, I think it should return the banpass filtered image/linescan (not the SpanROIs). I would suggest to still allow to set this. But if it is too complicated, then don't worry.

As for the list of SpanROI as arguments, that would be great to be added!

jlaehne:
@LMSC-NTappy check out this PR

jlaehne:
> With some guidance I would be happy to help adapt the code to be incorporated into hyperspy.

I still think an upstream contribution would be great. To have the functionality available more quickly, a way forward could be to merge it in LumiSpy and make a release. Once it is also available in HyperSpy, we could deprecate it on the LumiSpy side.

> I think knowing how best to handle the various shapes of data will be the challenge here.

Currently you have a span-ROI, but in a 2D map one would have rectangular ROIs to select parts of the data for the filtering. Would be surely nice to have that as well. Actually, once we integrate more functionality for Streak-Camera-Images, it would be helpful to have 2D signal support for LumiSpy as well.

@0Hughman0 could you solve the merge conflicts?

0Hughman0:
Hi,

Currently rushing to write up my PhD!

Would be happy to do this, but will have to wait until likely April.

LMSC-NTappy:
> @LMSC-NTappy check out this PR

Will do!

lumispy/utils/analysis.py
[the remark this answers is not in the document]
LMSC-NTappy: I agree with @jordiferrero and @jlaehne on this :)

lumispy/utils/analysis.py
LMSC-NTappy: I have nothing against having different functions for 0NavD, 1NavD, 2NavD signals but I let you decide :)

lumispy/utils/analysis.py
LMSC-NTappy: I guess here nothing really prevents invoking twice `plot_span_map` on the same data, no? I guess the interactivity is lost in that case

lumispy/utils/analysis.py
LMSC-NTappy: > This has made me realise that we should probably provide a spans argument which allows a user to provide a list of SpanROIs to use to slice your data, with ready defined ranges. This is the sort of thing I do often - use interactive ROIs to explore and find a desired ROI, and then overwrite the previous ROI definition to make the default position the one I just found... if that makes sense!

This! I think it's very important also for sharing notebooks etc. Personally I would be in favor of adding a list of tuples, something like `spans=[(beg0,end0),(beg1,end1)]` and not already defined roi objects, as the target audience of this feature, if I am not mistaken, are primarily people who do not want to bother learning how roi objects work am I right?

LMSC-NTappy (commented):
Very clean implementation and definitely a useful feature. 

I guess the few comments will be solved rather quickly, in my opinion only the question of module placement (plot vs analysis) is important as changing this later will be a pain, and the question of adding default arguments.

Other points of concern (>3 ranges and navigation dimensions) do not need to prevent a merge in my opinion and can be included later as improvements.

jlaehne:
Any time to upstream this to HyperSpy, could possibly make it into the v2.0 release scheduled for September?

0Hughman0:
Hiya,

Sorry for the delay, thanks for your patience!

I'll have a go at getting this going swiftly.

I'm trying to find when the release is scheduled for 2.0, but no luck? I can hopefully sort in the next few days.

I guess I should make a new PR on the hyperspy repo?

Change's I'll make:

- [ ] Add the ability to pass SpanROI not just ranges
- [ ] Move into the plot module
- [ ] Try and up the test coverage!
- [ ] Fix conflicts

0Hughman0:
This PR has been moved to one for hyperspy:

https://github.com/hyperspy/hyperspy/pull/3224

PR #156, opened 2022-10-22
Fix slicing of energy/wavenumber signal with isig

### LINKED ISSUE #155: Slicing of energy/wavenumber signal with isig fails
#### Describe the bug
Error occurs when slicing a signal with `isig[x:y]` that was converted using the lumispy functions `to_eV()` or `to_invcm()`.

#### To Reproduce
```
import hyperspy.api as hs
S = hs.signals.Signal1D(np.arange(100),axes=[{'axis': np.arange(100)+300}])
S.set_signal_type("CL")
S.to_eV(inplace=True)
S.isig[3.251:4.052]
```

Leads to `TypingError`:

```
TypingError: Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<function diff at 0x7fad54e2c310>) found for signature:
 
 >>> diff(array(float64, 1d, A))
 
There are 2 candidate implementations:
  - Of which 2 did not match due to:
  Overload in function 'np_diff_impl': File: numba/np/arraymath.py: Line 3684.
    With argument(s): '(array(float64, 1d, A))':
   Rejected as the implementation raised a specific error:
     TypingError: Failed in nopython mode pipeline (step: nopython frontend)
   - Resolution failure for literal arguments:
   reshape() supports contiguous array only
   - Resolution failure for non-literal arguments:
   reshape() supports contiguous array only
   
   During: resolving callee type: BoundFunction(array.reshape for array(float64, 1d, A))
   During: typing of call at /usr/lib/python3.10/site-packages/numba/np/arraymath.py (3702)
   
   
   File "../../../../usr/lib/python3.10/site-packages/numba/np/arraymath.py", line 3702:
       def diff_impl(a, n=1):
           <source elided>
           # To make things easier, normalize input and output into 2d arrays
           a2 = a.reshape((-1, size))
           ^

  raised from /usr/lib/python3.10/site-packages/numba/core/typeinfer.py:1086

During: resolving callee type: Function(<function diff at 0x7fad54e2c310>)
During: typing of call at /home/jonas/medien/pdi/git/hyperspy/hyperspy/misc/array_tools.py (428)


File "../../medien/pdi/git/hyperspy/hyperspy/misc/array_tools.py", line 428:
def numba_closest_index_round(axis_array, value_array):
    <source elided>
    rtol = 1e-10
    machineepsilon = np.min(np.abs(np.diff(axis_array))) * rtol
    ^
```

#### Expected behavior
If the energy is set up manually it works
```
import hyperspy.api as hs
import lumispy as lum
S1 = hs.signals.Signal1D(np.arange(100),axes=[{'axis': lum.nm2eV((np.arange(100)+300)[::-1])}])
S.set_signal_type("CL")
S.isig[3.251:4.052]
```

#### Additional context
Fix will be submitted, forcing the dtype in `axis2eV` and `axis2invcm`.

jlaehne:
### Description of the change
Fixes #155.

`.isig[]` was failing on signals converted using `to_eV()` or `to_invcm()`.

For tests to run, deprecated `--use-feature=2020-resolver` had to be removed from `PIP_ARGS` in `workflows/tests.yml`

### Progress of the PR
- [n/a] docstring updated (if appropriate),
- [n/a] update user guide (if appropriate),
- [x] added tests,
- [x] ready for review.

```
import hyperspy.api as hs
S = hs.signals.Signal1D(np.arange(100),axes=[{'axis': np.arange(100)+300}])
S.set_signal_type("CL")
S.to_eV(inplace=True)
S.isig[3.251:4.052]
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,7 +9,7 @@ jobs:
     timeout-minutes: 30
     env:
       MPLBACKEND: agg
-      PIP_ARGS: --upgrade --use-feature=2020-resolver -e
+      PIP_ARGS: --upgrade -e
       PYTEST_ARGS: --pyargs lumispy
       PYTEST_ARGS_COVERAGE: 
     strategy:
--- a/lumispy/utils/axes.py
+++ b/lumispy/utils/axes.py
@@ -87,10 +87,10 @@ def axis2eV(ax0):
         raise AttributeError("Signal unit is already eV.")
     # transform axis, invert direction
     if ax0.units == "µm":
-        evaxis = nm2eV(1000 * ax0.axis)[::-1].astype("float64")
+        evaxis = nm2eV(1000 * ax0.axis)[::-1].astype("float")
         factor = 1e3  # correction factor for intensity
     else:
-        evaxis = nm2eV(ax0.axis)[::-1].astype("float64")
+        evaxis = nm2eV(ax0.axis)[::-1].astype("float")
         factor = 1e6
     axis = DataAxis(axis=evaxis, name="Energy", units="eV", navigate=False)
     return axis, factor
@@ -149,10 +149,10 @@ def axis2invcm(ax0):
         raise AttributeError(r"Signal unit is already cm$^{-1}$.")
     # transform axis, invert direction
     if ax0.units == "µm":
-        invcmaxis = nm2invcm(1000 * ax0.axis)[::-1].astype("float64")
+        invcmaxis = nm2invcm(1000 * ax0.axis)[::-1].astype("float")
         factor = 1e4  # correction factor for intensity
     else:
-        invcmaxis = nm2invcm(ax0.axis)[::-1].astype("float64")
+        invcmaxis = nm2invcm(ax0.axis)[::-1].astype("float")
         factor = 1e7
     axis = DataAxis(
         axis=invcmaxis, name="Wavenumber", units=r"cm$^{-1}$", navigate=False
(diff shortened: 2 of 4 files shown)


jlaehne:
@mikelgr11 you had observed the problem reported in #155, which should fixed with this PR.

.github/workflows/tests.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     timeout-minutes: 30
     env:
       MPLBACKEND: agg
-      PIP_ARGS: --upgrade --use-feature=2020-resolver -e
```
jordiferrero, line 12: Can you explain to me a bit more what this change means?

jordiferrero (approved):

lumispy/utils/axes.py
```diff
@@ -149,10 +149,10 @@ def axis2invcm(ax0):
         raise AttributeError(r"Signal unit is already cm$^{-1}$.")
     # transform axis, invert direction
     if ax0.units == "µm":
-        invcmaxis = nm2invcm(1000 * ax0.axis)[::-1]
+        invcmaxis = nm2invcm(1000 * ax0.axis)[::-1].astype("float64")
```
jordiferrero, line 152: Wouldn't `float` be better (to account for x32bit computers too)?

.github/workflows/tests.yml
jlaehne: The `--use-feature=2020-resolver` option is leading to a failure of the workflow. As far as I can see it is deprecated and as the workflows in other hyperspy-related projects don't use it, I think we can do without. I guess, it comes from some template that you used when setting up the tests.

lumispy/utils/axes.py
jlaehne: Indeed, just took what the working format was locally - but the more general one would indeed be better.

PR #168, opened 2022-12-29
Add centroid functionality

jordiferrero:
### Description of the change
Added a centroid/center of mass functionality to analyze spectra.
- Added center of mass `com` utility
- Added an `centroid` method in the `LumiSpectrum` class

### Progress of the PR
- [x] docstring updated,
- [x] update user guide,
- [x] added tests,
- [x] ready for review.

### Minimal example of the new feature
```python
import lumispy as lum
s = lum.signals.LumiSpectrum([[[1,2,3,2,1,0]]*3]*4)
# s --> LumiSpectrum <3,4|6>

ax = s.axes_manager.signal_axes[0]
ax.offset = 200
ax.scale = 100

com = s.centroid()
# com --> LumiSpectrum <3,4|1>
# com.data[0,0] = 400.
```


jlaehne:
For a cleaner history, could you squash all the `fix` commits?
See https://hyperspy--3064.org.readthedocs.build/en/3064/dev_guide/git.html#keeping-the-git-history-clean

lumispy/utils/signals.py
```diff
@@ -0,0 +1,35 @@
...
+        The centroid of the spectrum.
+
+    Examples
+    --------
+    # Assume we have a spectrum with wavelengths and intensities
+    >>> wavelengths = [200, 300, 400, 500, 600, 700]
+    >>> intensities = [1, 2, 3, 2, 1, 0]
+
+    >>> center_of_mass = com(wavelengths, intensities)
+    >>> print(center_of_mass)  # Outputs: 400.0
```
jlaehne, line 25: Did you check that code? Does printing of a signal yield the data value?

jlaehne:
Nice functionality! I added a few first comments after having a quick look.

Do you have an idea, why the azure pipelines are failing?

As it is implemented (with a sum), the function works for uniform data axes only! That should be documented/checked. Alternatively, implement a version for non-uniform axes?

lumispy/utils/signals.py
jordiferrero: That's not a signal. That's just the utils function.

setup.py
```diff
@@ -53,14 +53,18 @@
     packages=find_packages(),
     # adjust the tabbing
     install_requires=[
-        "numpy",
```
ericpre, line 56: Why is this required?

lumispy/utils/signals.py
```diff
@@ -0,0 +1,35 @@
...
+    >>> center_of_mass = com(wavelengths, intensities)
+    >>> print(center_of_mass)  # Outputs: [400.0]
+    """
+    if len(spectrum_intensities) != len(wavelengths):
+        raise ValueError(
+            f"The length of the spectrum array {len(spectrum_intensities)} must match the length of the wavelength array {len(wavelengths)}."
+        )
+
+    wavelengths = np.array(wavelengths)
+    intensities = np.array(spectrum_intensities)
```
ericpre, line 33: According to the docstring, it should already be an array, is it necessary then?

lumispy/utils/signals.py
```diff
@@ -0,0 +1,35 @@
...
+    center_of_mass = np.sum(wavelengths * intensities) / np.sum(intensities)
+    return [center_of_mass]
```
ericpre, line 35: Should it return a float? If not, the docstring needs to be updated accordingly.

lumispy/signals/luminescence_spectrum.py
```diff
@@ -562,6 +586,53 @@ def px_to_nm_grating_solver(
...
+        """
+        Finds the centroid (center of mass) of a peak in the spectrum from the wavelength (or pixel number) and the intensity at each pixel value. It basically represents a "weighted average" of the peak.
+
+        Notes
+        -------
+        This function only works for a single peak. If you have multiple peaks, slice the signal beforehand or use the slice parameter.
+
+        TODO
+        -------
+        Implement this function for multiple peaks (npeaks = 2) by finding the top 2 peaks from mean spectrum and then returning a signal with 2 com.
```
ericpre, line 599: There is no `TODO` section in numpydoc (https://numpydoc.readthedocs.io/en/latest/format.html).

lumispy/signals/luminescence_spectrum.py
```diff
@@ -562,6 +586,53 @@ def px_to_nm_grating_solver(
         "\n", "\n\t"
     )
 
+    def centroid(self, slice=None, npeaks=1):
```
ericpre, line 589: If this isn't used, there is no need to create this parameter. It should be added when it is implemented.

lumispy/signals/luminescence_spectrum.py
```diff
@@ -562,6 +586,53 @@ def px_to_nm_grating_solver(
...
+        if npeaks > 1 or npeaks < 1:
+            raise NotImplementedError("Centroid only works for one peak.")
+
+        wavelengths = s.axes_manager.signal_axes[0].axis
+        center_of_mass = s.map(com, wavelengths=wavelengths, inplace=False)
+
+        # Transfer axes metadata
+        ax = center_of_mass.axes_manager.signal_axes[0]
+        ax.units = s.axes_manager.signal_axes[0].units
+        ax.name = s.axes_manager.signal_axes[0].name
```
ericpre, line 633: Why is it necessary? The returned signal should have a signal dimension of 0, i .e. at each navigation index, there is a scalar value.

setup.py
jordiferrero: No idea. Will remove it.

lumispy/signals/luminescence_spectrum.py
jordiferrero: I wanted to avoid this but you are right. I will save the units/name of the signal axes (that disappears) in the title of the BaseSignal that stays.

jordiferrero:
I have now added all your feedback.
I have included support for any axis (uniform and non-uniform). I have tried to squash the commits but I don't seem to make it work... sorry. I also tried to fix Azure tests failing but is seems to be a problem not linked to this PR. I will open a new issue on that matter.
Let me know what you think now. Thanks

setup.py
ericpre: I meant the pinning on `numpy` that you introduce in this PR. `numpy` and `scipy` are dependencies, so they should stay here, but the pinning needed to be removed, as there are no reasons to pin them.

azure-pipelines.yml
```diff
@@ -50,7 +50,7 @@ steps:
 - bash: |
     source activate $ENV_NAME
     # Install package using pip
-    pip install . --no-deps
+    conda install .
```
ericpre, line 53: What is the reason for this change?

jordiferrero: In the azure-pipelines.yml it says that it should test for using conda and only conda prompts.
It used to use pip, which does not use conda. That's why I changed it. But I can change it back as it is not part of this PR.

ericpre: > In the azure-pipelines.yml it says that it should test for using conda and only conda prompts. It used to use pip, which does not use conda. That's why I changed it. But I can change it back as it is not part of this PR.

This refers to dependencies/environment. Generally it is good practice not to mix different things in PR. If you think, this needs to be changed, it should be in a separate PR.

jordiferrero: I agree. I have now reverted this change. Ready for review once more.

ericpre (commented):
In HyperSpy, there is a [`find_peak_ohaver` ](https://hyperspy.readthedocs.io/en/latest/api/hyperspy._signals.signal1d.html#hyperspy._signals.signal1d.Signal1D.find_peaks1D_ohaver) and I think that it would be useful to merge it with the function of this PR into a generic `find_peak()`, where method could be `center_of_mass` or `ohaver`?

jordiferrero:
> In HyperSpy, there is a [`find_peak_ohaver` ](https://hyperspy.readthedocs.io/en/latest/api/hyperspy._signals.signal1d.html#hyperspy._signals.signal1d.Signal1D.find_peaks1D_ohaver) and I think that it would be useful to merge it with the function of this PR into a generic `find_peak()`, where method could be `center_of_mass` or `ohaver`?

I am well aware of `find_peaks_ohaver()`. I don't think `centroid` and `o_haver` are similar enough to be merged together. Centroid does not find peaks, it finds the center of mass of a peak, once you already know it exists... 
In fact, an implementation I wanted to add in the future is to be able to get centroids for spectra that have more than one peak (currently, you need to crop the signal so only one peak exists to get something physical). For this future implementation, `peak_o_haver` will be used to find the main peaks, then centroids for each peak will be calculated. That's why I don't think these 2 functions should be merged.
Also, centroid analysis is very relevant for luminescence studies, that's why I'd prefer it to be in the LumiSpy package.

lumispy/signals/luminescence_spectrum.py
```diff
@@ -562,6 +586,49 @@ def px_to_nm_grating_solver(
         "\n", "\n\t"
     )
 
+    def centroid(self, slice=None, **kwargs):
+        """
+        Finds the centroid (center of mass) of a peak in the spectrum from the wavelength (or pixel number) and the intensity at each pixel value. It basically represents a "weighted average" of the peak.
+
+        Notes
```
jlaehne, line 593: As in the `rst` files, same length of underlining as the preceding title

doc/source/user_guide/utilities.rst
```diff
@@ -113,9 +113,51 @@ index of air.
...
 `Horiba Scientific <https://horiba.com/uk/scientific/products/optics-tutorial/wavelength-pixel-position>`_.
+
+
+.. _centroid:
+
+Calculating the centroid of a spectrum (centre of mass)
+==============================
+
+
+The function :py:meth:`~.signals.luminescence_spectrum.LumiSpectrum.centroid` (based on the utility function :py:func:`~.utils.signals.com`) is an alternative to finding the max intensity of a peak.
```
jlaehne, line 130: Use line breaks after about 80 characters for better readability of the raw text also in the documentation.
 "max" -> "maximum"

setup.py
jlaehne: still open

jlaehne:
> I have included support for any axis (uniform and non-uniform).
  I'll try to find some time to test it on some data the next days.

lumispy/signals/luminescence_spectrum.py
```diff
@@ -562,6 +586,49 @@ def px_to_nm_grating_solver(
         "\n", "\n\t"
     )
 
+    def centroid(self, slice=None, **kwargs):
+        """
+        Finds the centroid (center of mass) of a peak in the spectrum from the wavelength (or pixel number) and the intensity at each pixel value. It basically represents a "weighted average" of the peak.
```
jlaehne, line 591: Now not restricted to wavelength!

lumispy/utils/signals.py
```diff
@@ -0,0 +1,74 @@
+import numpy as np
+from hyperspy.axes import FunctionalDataAxis
+from scipy.ndimage import center_of_mass
+from scipy.interpolate import interp1d
+
+
+def com(spectrum_intensities, signal_axis, **kwargs):
+    """Finds the centroid (center of mass) of a peak in the spectrum based from the intensity at each pixel value and a its respective signal axis.
```
jlaehne, line 8:
```suggestion
    """Finds the centroid (center of mass) of a peak in the spectrum based from the intensity at each pixel value and its respective signal axis.
```

jlaehne (approved):

jlaehne:
Thanks @jordiferrero for this great work. Just tried it on a sample dataset, which is as if it was made for this functionality. And it works both for wavelength and energy axis.

When the Jacobian is deactivated, the results in wavelength and energy range agree, as they should. When the Jacobian is activated, there is a notable shift of the value in energy (when back-converted to wavelength). As for fitting, these kind of routines should be performed in the energy scale! And we should add a note to the user guide (by the way, we still have to write the corresponding paragraph in the fitting chapter of the user guide, it is still a todo-note).

Apart from this note and the minor comments above (mainly about formatting in rest format and docstrings), this PR is ready to be merged in my view. I'm happy to do so once the last things are done.

However, being at it we should also create a corresponding maximum function that finds the maximum of a peak - e.g. by taking the first derivative and then finding zero-crossings.

lumispy/signals/luminescence_spectrum.py
jlaehne: Line wrapping in docstrings and documentation would be good.

ericpre:
> > In HyperSpy, there is a [`find_peak_ohaver` ](https://hyperspy.readthedocs.io/en/latest/api/hyperspy._signals.signal1d.html#hyperspy._signals.signal1d.Signal1D.find_peaks1D_ohaver) and I think that it would be useful to merge it with the function of this PR into a generic `find_peak()`, where method could be `center_of_mass` or `ohaver`?
> 
> I am well aware of `find_peaks_ohaver()`. I don't think `centroid` and `o_haver` are similar enough to be merged together. Centroid does not find peaks, it finds the center of mass of a peak, once you already know it exists... In fact, an implementation I wanted to add in the future is to be able to get centroids for spectra that have more than one peak (currently, you need to crop the signal so only one peak exists to get something physical). For this future implementation, `peak_o_haver` will be used to find the main peaks, then centroids for each peak will be calculated. That's why I don't think these 2 functions should be merged. Also, centroid analysis is very relevant for luminescence studies, that's why I'd prefer it to be in the LumiSpy package.

I think both approaches are valid and subjective and in this kind of situation, I would favour the approach which provide the simplest and more consistent API. Once thing that I find inconvenient is to have functions or methods which does similar things because it makes the namespace more busy and quite often it is not obvious to figure out what are the difference between them.

lumispy/signals/luminescence_spectrum.py
```diff
@@ -562,6 +586,49 @@ def px_to_nm_grating_solver(
...
+        Notes
+        -------
+        This function only works for a single peak. If you have multiple peaks, slice the signal beforehand or use the slice parameter.
+
+        TODO: Implement this function for multiple peaks (with the npeaks parameter) by finding the top 2 peaks from mean spectrum and then returning a signal with 2 com.
+
+        Parameters
+        -------
+        slice : tuple (2)
+            An tuple representing the indices of the wavelength (start index, end index) where the peak is located. If the tuple contains int, it slices on index. Ig the tuple contains float, it slices on signal units (defualt hyperspy s.inav[:] functionality).
```
jlaehne, line 602:
```suggestion
            A tuple representing the indices of the wavelength (start index, end index) where the peak is located. If the tuple contains int, it slices on index. If the tuple contains float, it slices on signal units (default HyperSpy s.inav[:] functionality).
```

lumispy/signals/luminescence_spectrum.py
```diff
@@ -562,6 +586,49 @@ def px_to_nm_grating_solver(
...
+
+        Notes
+        -------
+        This function only works for a single peak. If you have multiple peaks, slice the signal beforehand or use the slice parameter.
+
+        TODO: Implement this function for multiple peaks (with the npeaks parameter) by finding the top 2 peaks from mean spectrum and then returning a signal with 2 com.
+
+        Parameters
+        -------
+        slice : tuple (2)
```
jlaehne, line 601: In HyperSpy this parameter is in fact called `signal_range`, see e.g. [remove_background](https://hyperspy.readthedocs.io/en/latest/api/hyperspy._signals.signal1d.html#hyperspy._signals.signal1d.Signal1D.remove_background), and we should be consistent there. HyperSpy allows for "interactive" slicing, which would be a worthwile addition.

jlaehne:
> I think both approaches are valid and subjective and in this kind of situation, I would favour the approach which provide the simplest and more consistent API. Once thing that I find inconvenient is to have functions or methods which does similar things because it makes the namespace more busy and quite often it is not obvious to figure out what are the difference between them.

I am not so sure about this.

1. `find_peaks1D_ohaver` would have to be renamed as the centroid functionality definitely is something different than an OHaver routine.

2. Both functionalities would take quite different parameters, which makes the docstring much more complicated.

So I tend to opt for the separate implementation.

ericpre:
I am fine with both approaches - I am just being annoying to keep namespace and API tidy! :)

lumispy/signals/luminescence_spectrum.py
jordiferrero: I have changed the name of the parameter to match the standards.
However, adding an "interactive" functionality would require a significant amount of extra coding I don't think would be worth it. If you still think it is necessary, we can delay this PR more and I can attempt to add it in the future. I'd simply leave it as is and maybe add it in the future as an additional feature.

jordiferrero:
I have now implemented the small formatting you requested. I have also moved the documentation to the signals page.
I think if you prefer to call this function `find_peaks1D_centroid` instead of `centroid`, I am happy to change it to that. I find that a bit misleading as to me this method is not finding peaks... it finds mass density, which does not necessarily relate to peaks but shoulders in emission and asymmetry detection... Let me know what you think, else I am happy for you to do the last review.

doc/source/user_guide/utilities.rst
```diff
@@ -9,7 +9,7 @@ This section summarizes various useful functions implemented in LumiSpy.
 .. _join_spectra:
 
 Join spectra
-============
+==============================
```
jlaehne, line 12:
```suggestion
==========
```

doc/source/user_guide/utilities.rst
```diff
@@ -30,7 +30,7 @@ signals in the range of +/- 50 pixels around the centre of the overlapping regio
 .. _spectral_map_utils:
 
 Utilities for spectral maps
-===========================
+==============================
```
jlaehne, line 33:
```suggestion
===========================
```

doc/source/user_guide/utilities.rst
```diff
@@ -47,7 +47,7 @@ HyperSpy <signal.indexing>`.
 .. _unit_conversion:
 
 Unit conversion
-===============
+==============================
```
jlaehne, line 50:
```suggestion
===============
```

lumispy/signals/luminescence_spectrum.py
```diff
@@ -562,6 +586,61 @@ def px_to_nm_grating_solver(
...
     )
 
+    def centroid(self, signal_range=None, **kwargs):
+        """
+        Finds the centroid (center of mass) of a peak in the spectrum from
+        the wavelength (or pixel number) and the intensity at each pixel
+        value. It basically represents a "weighted average" of the peak.
+
+        Notes
+        -------
```
jlaehne, line 596:
```suggestion
        -----
```

lumispy/signals/luminescence_spectrum.py
```diff
@@ -562,6 +586,61 @@ def px_to_nm_grating_solver(
...
+        This function only works for a single peak. If you have multiple
+        peaks, slice the signal beforehand or use the signal_range parameter.
+
+        TODO: Implement this function for multiple peaks (with the npeaks
+        parameter) by finding the top 2 peaks from mean spectrum and then
+        returning a signal with 2 com.
+
+        Parameters
+        -------
```
jlaehne, line 605:
```suggestion
        ----------
```

lumispy/utils/signals.py
```diff
@@ -0,0 +1,78 @@
+import numpy as np
+from hyperspy.axes import FunctionalDataAxis
+from scipy.ndimage import center_of_mass
+from scipy.interpolate import interp1d
+
+
+def com(spectrum_intensities, signal_axis, **kwargs):
+    """Finds the centroid (center of mass) of a peak in the spectrum based
+    from the intensity at each pixel value and its respective signal axis.
```
jlaehne, line 9:
```suggestion
    from the intensity at each pixel value and its respective signal axis.

```

lumispy/utils/signals.py
```diff
@@ -0,0 +1,78 @@
...
+    Parameters
+    ----------
+    spectrum_intensities : array
+        An array with the intensities of the spectrum.
+    signal_axis: hyperspy.axes.BaseDataAxis subclass
+        A HyperSpy signal axis class containing an array with the wavelength/
+        energy for each intensity/signal value.
+    kwargs : dictionary
+        For the scipy.interpolate.interp1d function.
```
jlaehne, line 18:
```suggestion
        For the scipy.interpolate.interp1d function.

```

setup.py
```diff
@@ -53,14 +53,16 @@
     packages=find_packages(),
     # adjust the tabbing
     install_requires=[
```
jlaehne, line 55:
```suggestion
    install_requires=[
        "numpy",
        "scipy",
```

lumispy/utils/signals.py
```diff
@@ -0,0 +1,78 @@
...
+        com_val = signal_axis._function(x, **kwargs)
+
+    elif hasattr(signal_axis, "axis"):
+        # Calculate value interpolating between index_com 0 and 1
+        com_val = _interpolate_signal(signal_axis.axis, index_com)
+    elif type(signal_axis) in (list, np.ndarray, tuple):
+        # Check for dimensionality
+        if len(spectrum_intensities) != len(signal_axis):
+            raise ValueError(
+                f"The length of the spectrum array {len(spectrum_intensities)} must match the length of the wavelength array {len(signal_axis)}."
```
jlaehne, line 71:
```suggestion
                f"The length of the spectrum array {len(spectrum_intensities)} must match "
                "the length of the wavelength array {len(signal_axis)}."
```

lumispy/signals/luminescence_spectrum.py
```diff
@@ -562,6 +586,61 @@ def px_to_nm_grating_solver(
...
+            each navigation pixel.
+        """
+        if signal_range:
+            if type(signal_range) != tuple:
+                raise TypeError(
+                    "The `signal_range` parameter must be a tuple of length 2."
+                )
+            if len(signal_range) != 2:
+                raise ValueError(
+                    f"The `signal_range` parameter must be a tuple of length 2. You passed a tuple of length {len(signal_range)}"
```
jlaehne, line 629:
```suggestion
                    f"The `signal_range` parameter must be a tuple of length 2. "
                    "You passed a tuple of length {len(signal_range)}."
```

doc/source/user_guide/signal_tools.rst
```diff
@@ -97,6 +97,46 @@ default value ``factor=0.5`` returns the full width at half maximum (FWHM).
...
+
+    >>> com = s.centroid()
+    >>> com
+    BaseSignal <2,3|>
+    >>> com.data[0,0] 
+    400.
+
+.. Note::
+
+    This function only works for a single peak. If you have multiple peaks, slice the signal beforehand or use the slice parameter (which follows the ``s.isig[:]`` convention).
```
jlaehne, line 135:
```suggestion
    This function only works for a single peak. If you have multiple peaks,
    slice the signal beforehand or use the slice parameter (which follows the
    ``s.isig[:]`` convention).
```

doc/source/user_guide/signal_tools.rst
```diff
@@ -97,6 +97,46 @@ default value ``factor=0.5`` returns the full width at half maximum (FWHM).
...
+
+.. Note::
+
+    This function is good at identifying non-symmetric peaks with shoulders. Such changes would not be reflected in the peak maximum.
```
jlaehne, line 139:
```suggestion
    The :ref:`jacobian` may affect the shape, in particular of broader peaks.
    It is therefore highly recommended to convert luminescence spectra from
    wavelength to the :ref:`energy axis <energy_axis>` prior to determining
    the centroid to determine the true emission energy.
    See e.g. [Wang]_ and [Mooney]_.
```

doc/source/user_guide/signal_tools.rst
```diff
@@ -97,6 +97,46 @@ default value ``factor=0.5`` returns the full width at half maximum (FWHM).
...
+
+
+The function :py:meth:`~.signals.luminescence_spectrum.LumiSpectrum.centroid` (based on the utility function :py:func:`~.utils.signals.com`) is an alternative to finding the max intensity of a peak.
+It finds the centroid (center of mass) of a peak in the spectrum from the wavelength (or pixel number) and the intensity at each pixel value. It basically represents a "weighted average" of the peak as such:
+
+.. math::
+
+    com = \frac{\sum{x_i I_i}}{\sum{I_i}},
+
+where :math:`x_i` is the wavelength (or pixel number) at which the intensity of the spectrum :math:`I_i` is measured.
```
jlaehne, line 113:
```suggestion
where :math:`x_i` is the wavelength (or pixel number) at which the
intensity of the spectrum :math:`I_i` is measured.
```

doc/source/user_guide/signal_tools.rst
```diff
@@ -97,6 +97,46 @@ default value ``factor=0.5`` returns the full width at half maximum (FWHM).
...
+
+This function also works for non-linear axes. For the :external:py:class:`hyperspy.axes.FunctionalDataAxis`, the centroid is extrapolated based on the function used to create the non-uniform axis. For :external:py:class:`hyperspy.axes.DataAxis`, a linear interpolation between the axes points at the center of mass is assumed, but this behaviour can be changed with the `kwargs` of :external:py:meth:`scipy.interpolate.interp1d` function.
```
jlaehne, line 115:
```suggestion
This function also works for non-linear axes. For the
:external:py:class:`hyperspy.axes.FunctionalDataAxis`, the centroid is extrapolated
based on the function used to create the non-uniform axis. For
:external:py:class:`hyperspy.axes.DataAxis`, a linear interpolation between the
axes points at the center of mass is assumed, but this behaviour can be changed
with the `kwargs` of :external:py:meth:`scipy.interpolate.interp1d` function.
```

doc/source/user_guide/signal_tools.rst
```diff
@@ -97,6 +97,46 @@ default value ``factor=0.5`` returns the full width at half maximum (FWHM).
     >>> width = s.estimate_peak_width(factor=0.3)
 
 
+.. _centroid:
+
+Calculating the centroid of a spectrum (centre of mass)
+-------------------------------------------------------
+
+
+The function :py:meth:`~.signals.luminescence_spectrum.LumiSpectrum.centroid` (based on the utility function :py:func:`~.utils.signals.com`) is an alternative to finding the max intensity of a peak.
```
jlaehne, line 106:
```suggestion
The function :py:meth:`~.signals.luminescence_spectrum.LumiSpectrum.centroid`
(based on the utility function :py:func:`~.utils.signals.com`) is an alternative to
finding the position of the maximum intensity of a peak, useful in particular for
non-symmetric peaks with pronounced shoulders.
```

doc/source/user_guide/signal_tools.rst
```diff
@@ -97,6 +97,46 @@ default value ``factor=0.5`` returns the full width at half maximum (FWHM).
...
+It finds the centroid (center of mass) of a peak in the spectrum from the wavelength (or pixel number) and the intensity at each pixel value. It basically represents a "weighted average" of the peak as such:
```
jlaehne, line 107:
```suggestion
It finds the centroid (center of mass) of a peak in the spectrum from the signal axis
units (or pixel number) and the intensity at each pixel value. It basically represents a
"weighted average" of the peak as such:
```

jlaehne (approved):

lumispy/signals/luminescence_spectrum.py
jlaehne: Yes, lets keep it as a potential future feature. Just noticed that similar HyperSpy functions have that option - though I personally do not tend to use it much, I know some other people might appreciate.

jlaehne:
Thanks @jordiferrero for this nice feature.

I corrected a little more formatting, mainly in the documentation before merging.

PR #176, opened 2023-01-22
Fix list formatting

jlaehne:
### Description of the change
Fixing list in `signal_tools.rst`.

### Progress of the PR
- [ ] Change implemented (can be split into several points),
- [ ] ready for review.

### DIFF
--- a/doc/source/user_guide/signal_tools.rst
+++ b/doc/source/user_guide/signal_tools.rst
@@ -53,14 +53,14 @@ Peak identification
 HyperSpy provides functions to find the positions of maxima or minima in a
 dataset:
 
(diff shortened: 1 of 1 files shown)


jlaehne:
Hm, anyone an idea, why the list in https://lumispy--176.org.readthedocs.build/en/176/user_guide/signal_tools.html (peak identification) is not properly rendered, while it was OK on my local preview?

jordiferrero:
Actually, all bullet lists are broken now in the documentation. It seems that it all started in #168... 
However, the only thing that changed linked to documentation is be in the `setup.py` where better formatting was changed. However, I have tried to revert this change in #178 and nothing has changed (lists still broken). So I am not sure.
There is nothing else I can see obvious...

PR #179, opened 2023-02-02
Speed up centroid function (remove interpolation)

jordiferrero:
### Description of the change
I realised I had overengineered the `com` utils function. I have now simplified it with a huge speed improvement. It does not rely on `scipy` interpolation anymore.

### Progress of the PR
- [x] ready for review.

### DIFF
--- a/lumispy/utils/signals.py
+++ b/lumispy/utils/signals.py
@@ -1,7 +1,5 @@
 import numpy as np
 from hyperspy.axes import FunctionalDataAxis
-from scipy.ndimage import center_of_mass
-from scipy.interpolate import interp1d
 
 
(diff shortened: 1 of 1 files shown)


jlaehne:
Indeed, the function runs through about 6-times faster on a map of mine. For the uniform axis, the results are the same - but when converting to non-uniform axes, the results are not the same as for the old implementation. In particular, for `jacobian=False` the centroid in the old implementation is the same in energy and wavelength scale, which makes sense. In the new implementation, they are not the same.

I think the weighted average that you calculate is correct for the equidistant scale, but for non-uniform axes you still would need interpolation if you want to use that formula.

jlaehne:
Continued in #261

PR #183, opened 2023-03-15
Release_v0.2.2

jlaehne:
### Description of the change
I'd like to push a patch release before the tutorial at the CL-EBIC workshop next week.

### Progress of the PR
- [x] bumped version,
- [X] updated CHANGELOG.md,
- [X] ready for review.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,8 +6,10 @@ All notable changes to this project will be documented in this file.
 The format is based on `Keep a Changelog <https://keepachangelog.com/en/1.0.0/>`_,
 and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0.html>`_.
 
-UNRELEASED
-==========
+.. _changes_0.2.2:
+
+2023-03-15 - version 0.2.2
+==========================
 Changed
 -------
 - Use `GitHub code scanning (CodeQL)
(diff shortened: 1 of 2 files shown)


ericpre:
> I'd like to push a patch release before the tutorial at the CL-EBIC workshop next week.

Do you think that we should release a hyperspy patch release asap to avoid potential troubles with the random number generation changes in recent version of dask?

jlaehne:
> > I'd like to push a patch release before the tutorial at the CL-EBIC workshop next week.
> 
> Do you think that we should release a hyperspy patch release asap to avoid potential troubles with the random number generation changes in recent version of dask?

Haven't looked into the details, but sounds sensible and might not hurt.

jlaehne:
I just realized that for a map the new `centroid` function, similar to e.g. `valuemax` and `indexmax` in HyperSpy returns a BaseSignal with 2 navigation dimensions and 0D signal, which theoretically makes sense, but is inconvenient for plotting as it plots including the navigator, which makes little sense for 0D signals, and colormap cannot be changed.

In contrast, `estimate_peak_width` returns the transposed signal, thus Signal2D object without navigation dimensions, which is more convenient for plotting - and is okay in the sense that it is just an image of values for the different positions.

We should consider making that more consistent also in HyperSpy (e.g. with the 2.0 release), but if we release the `centroid` function here and then decide to change the behavior again, we unnecessarily break the API.

jlaehne:
I changed the `centroid` output type to `Signal2D`  in #184 and opened an issue for HyperSpy https://github.com/hyperspy/hyperspy/issues/3111

Issue #185, opened 2023-03-17
Fix test for `remove_spikes`


jlaehne:
Some recent update to HyperSpy or an upstream library breaks the test for automatic peak removal:

```
>       np.testing.assert_almost_equal(s3.data[0, 2, 29], 1, decimal=5)
E       AssertionError: 
E       Arrays are not almost equal to 5 decimals
E        ACTUAL: 2.000005802611809
E        DESIRED: 1
```

You can see the full output here: https://github.com/LumiSpy/lumispy/actions/runs/4451990112/jobs/7819241076

The smaller of the two artificial peaks is not caught on every run, if the `threshold` is `auto`.

I commented out line 70 and 85 in `test_cl_spectrum.py` to make the tests run through in #184.
Also I set `decimal=4` (instead of `5`) in all assertions as it was loosing accuracy on some runs.

So far, I have not found why the numerics suddenly seem to be less reliable.

In principle, the function should still run, but it seems to be less robust.

Any idea @jordiferrero or @ericpre

Attolight-NTappy:
Hello!

Just wanted to share that  the incriminated test runs fine with my configuration, so I don't think it is an upstream library modification issue:

lumispy 0.2.3.dev0 (main branch)
hyperspy 1.8.0.dev0 (both RELEASE_next_minor and RELEASE_next_patch branches)

```
============================= test session starts =============================
collecting ... collected 2 items

test_cl_spectrum.py::TestCLSpectrum::test__make_signal_mask 
test_cl_spectrum.py::TestCLSpectrum::test_remove_spikes 

======================== 2 passed, 3 warnings in 3.57s ========================

Process finished with exit code 0
PASSED       [ 50%]PASSED           [100%]
```

I have python 3.10.9 and numpy 1.23.5.

PR #187, opened 2023-03-23
fixed test

Attolight-NTappy:
Fixes the failing test in #185

### DIFF
--- a/lumispy/tests/signals/test_cl_spectrum.py
+++ b/lumispy/tests/signals/test_cl_spectrum.py
@@ -48,7 +48,6 @@ class TestCLSpectrum:
 
     def test_remove_spikes(self):
         s = CLSpectrum(np.ones((2, 3, 30)))
-        np.random.seed(np.random.randint(666666))
         s.add_gaussian_noise(1e-5)
         # Add three spikes
         s.data[1, 0, 1] += 2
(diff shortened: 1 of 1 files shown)


lumispy/tests/signals/test_cl_spectrum.py
```diff
@@ -48,11 +48,11 @@ def test__make_signal_mask(self):
 
     def test_remove_spikes(self):
         s = CLSpectrum(np.ones((2, 3, 30)))
-        np.random.seed(1)
+        np.random.seed(np.random.randint(666666))
```
ericpre, line 51: If I understand correctly, `np.random.randint` is still using the legacy random generator: https://numpy.org/doc/stable/reference/random/generated/numpy.random.randint.html. Would it make sense to use the new generator?
The legacy generator is not deprecated but it will most likely be at some point.

Attolight-NTappy: For what I understood the np.random.seed statement has no effect as the following add_gaussian_noise from hyperspy does not take it into account and applies its own seed using the generator object.

I updated the tests to remove this line

ericpre:
Does `remove_spikes` have any added functionalities in comparison to `Signal1D.spikes_removal_tool`? If not, is it worth keeping it?

Attolight-NTappy:
From what I see in the code, it mostly allows to choose whether it is done in place or not. 

(I just discover now that the test is still flaky)

jlaehne:
> Does `remove_spikes` have any added functionalities in comparison to `Signal1D.spikes_removal_tool`? If not, is it worth keeping it?

Indeed, apart from the additional `inplace` option, which should be easy to contribute upstream, and the possibility to build a `signal_mask` from a `luminescence_roi`, it is only a wrapper function with different defaults (`interactive=False`).

@jordiferrero, you contributed this function, what was the original motivation behind it? If we are just changing defaults of an existing function, it does not really make sense to keep it up.

jlaehne (approved):

PR #191, opened 2023-05-27
Update releasing_guide.md

jlaehne:
Some more details to releasing guide.

### Progress of the PR
- [x] ready for review.

### DIFF
--- a/releasing_guide.md
+++ b/releasing_guide.md
@@ -18,9 +18,13 @@ To publish a new LumiSpy release do the following steps:
 - Create a tag e.g. `git tag -a v0.1.1 -m "LumiSpy version 0.1.1"`
 - Push tag to user fork for a test run `git push origin v0.1.1`. Will run the release
   workflow without uploading to PyPi
-- Push tag to LumiSpy repository to trigger release `git push upstream v0.1.1`
-  (this triggers the GitHub action to create the sdist and wheel and upload to
-  PyPi automatically). :warning: this is a point of no return :warning:
+- Push tag to LumiSpy repository to trigger release `git push upstream v0.1.1`. 
+  :warning: this is a point of no return :warning:
+  - triggers the GitHub action to create the sdist and wheel and upload them to
+  PyPi automatically
+  - creates a Github Release
+  - creates a zenodo record and corresponding DOI (triggered by a new release)
+  _updates the current version on readthedocs to this release [we would need to set it to stable instead of latest]
   
 ## Post-release action
  


releasing_guide.md
```diff
@@ -18,9 +18,13 @@ To publish a new LumiSpy release do the following steps:
 - Create a tag e.g. `git tag -a v0.1.1 -m "LumiSpy version 0.1.1"`
 - Push tag to user fork for a test run `git push origin v0.1.1`. Will run the release
   workflow without uploading to PyPi
-- Push tag to LumiSpy repository to trigger release `git push upstream v0.1.1`
-  (this triggers the GitHub action to create the sdist and wheel and upload to
-  PyPi automatically). :warning: this is a point of no return :warning:
+- Push tag to LumiSpy repository to trigger release `git push upstream v0.1.1`. 
+  :warning: this is a point of no return :warning:
+  - triggers the GitHub action to create the sdist and wheel and upload them to
```
ericpre, line 23: Currently, the order is different:

1. create github release, which trigger the zenodo release
2. build wheel/sdist
3. upload to pypi

releasing_guide.md
```diff
@@ -18,9 +18,13 @@ To publish a new LumiSpy release do the following steps:
...
+  PyPi automatically
+  - creates a Github Release
+  - creates a zenodo record and corresponding DOI (triggered by a new release)
+  _updates the current version on readthedocs to this release [we would need to set it to stable instead of latest]
```
ericpre, line 27:
```suggestion
  - updates the current version on readthedocs to this release [we would need to set it to stable instead of latest]
```
This should be done automatically (to check), but it is not clear how `readthedocs.org` does it, i.e. update the `stable` version and it needs a specific setting. It is most likely trigger by a tag push rather the github release creation.

releasing_guide.md
ericpre, line 23: Maybe it would be better to create the github release after the packages are upload to pypi, in case the release workflow fails at some point, as this would to push again the tag without to worry about the Zenodo entry being created on the GitHub release creation.
I added some additional notes in https://github.com/hyperspy/rosettasciio/pull/126

jlaehne: You set it up in a different order for RSciIO - is there a particular advantage for that?

ericpre: Yes, see [comment below ](https://github.com/LumiSpy/lumispy/pull/191#discussion_r1208392335) and additional information section in https://github.com/hyperspy/rosettasciio/pull/126/files#diff-1ac981832aad6ef0ee37d78805ad59ce5cbd618646f64e363246b89acd57b312

releasing_guide.md
jlaehne: Currently, readthedocs is set to the `latest` version, thus the development version of the user guide. That was to get a first version running while it was not released/work in progress. However, as most topics are now covered, we should change it to `stable` sometime so that the docs correspond to the released version. Then this point would be valid and the commenting out can be removed.

releasing_guide.md
jlaehne: Valid point, maybe we should indeed adapt our release workflow to match the one of rosettasciio.

releasing_guide.md
ericpre: Oh, I thought this was a typo! 😅

jlaehne:
Replaced by https://github.com/LumiSpy/lumispy/pull/211

PR #202, opened 2023-12-21
Update supported python versions

jlaehne:
### Progress of the PR
- [x] ready for review.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,8 +4,8 @@ on: [push, pull_request]
 
 jobs:
   run_test_suite:
-    name: py${{ matrix.PYTHON_VERSION }}${{ matrix.LABEL }}
-    runs-on: ubuntu-latest
+    name: ${{ matrix.os }}-py${{ matrix.PYTHON_VERSION }}${{ matrix.LABEL }}
+    runs-on: ${{ matrix.os }}-latest
     timeout-minutes: 30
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -15,15 +15,21 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        PYTHON_VERSION: ['3.7', '3.9', '3.10']
+        os: [ubuntu, windows, macos]
```
ericpre, line 18: Add this in the name above:
```
name: ${{ matrix.os }}-py${{ matrix.PYTHON_VERSION }}${{ matrix.LABEL }}
```

.github/workflows/tests.yml
```diff
@@ -15,15 +15,21 @@ jobs:
...
+        PYTHON_VERSION: ['3.8', '3.12']
         PIP_SELECTOR: ['[tests]']
```
ericpre, line 20: The `PIP_SELECTOR` is missing for the `3.8` and `3.12` build, maybe there is some parsing issue and it needs a coma to make sure it is a list?

jlaehne:
@ericpre finally got it to work, the setting of 'bash' was important for it to run under windows as well.

In this context I realized that the hyperspy reusable workflow is running the coverage under all python versions and operating systems - wouldn't one run be enough?

PR #205, opened 2024-01-01
Axis conversion for TransientSpec

### LINKED ISSUE #204: Slicing TransientSpec
When slicing, summing, integrating signals of type `TransientSpec`, the resulting signal will be a `Signal1D`. It would be desirable to obtain signals of type `Transient` or `Luminescence` depending on what axis the function operates on.

jlaehne:
### Description of the change
Move axis conversion code to `CommonLuminescence` class to make it available to `LuminescenceTransientSpectrum` signals.

Handle inheritance for slicing of `LuminescenceTransientSpectrum`:
- Create 1D version of  `signal_type = 'TransientSpec'` (class `TransientSpectrumCasting`)
- If `TransientSpec` is initialized in 1D with `axes_manager[-1].units` being a time unit, switch to `LumiTransient` class, otherwise switch to `Luminescence` class

### Progress of the PR
- [x] move axis conversion code,
- [x] Implement workaround for https://github.com/hyperspy/hyperspy/issues/3294
- [x] Handle slicing of TransientSpec,
- [x] add section to user guide,
- [x] add tests for TransientSpec case,
- [x] added line to CHANGELOG.rst,
- [x] check formatting of changelog and user guid,
- [x] ready for review.


lumispy/signals/luminescence_transientspec.py
```diff
@@ -21,21 +21,47 @@
...
 
 
+class TransientSpectrumCasting(Signal1D, CommonLumi, CommonTransient):
+    """**1D signal class for casting reduced LumiTransientSpectrum to either Luminescence or Transient**"""
+
+    _signal_type = "TransientSpec"
+    _signal_dimension = 1
+
+    def __init__(self, *args, **kwargs):
+        if hasattr(self, "axes_manager") and self.axes_manager[-1].units in [
```
ericpre, line 39: Would it be worth using `pint` to check that the unit is of time dimension?

For example:
```python
import pint

ureg = pint.UnitRegistry()

assert ureg("ms").dimensionality == ureg("s").dimensionality

```

lumispy/hyperspy_extension.yaml
```diff
@@ -150,6 +150,17 @@ signals:
     lazy: True
     module: lumispy.signals.luminescence_transient
 
+  TransientSpectrumCasting:
+    signal_type: TransientSpec
```
ericpre, line 154: Use full word?
```suggestion
    signal_type: TransientSpectrum
```

lumispy/signals/luminescence_transientspec.py
```diff
@@ -21,21 +21,47 @@
...
+from hyperspy.signals import Signal1D, Signal2D
 from hyperspy._signals.lazy import LazySignal
 
+from lumispy.signals import LumiSpectrum, LumiTransient
 from lumispy.signals.common_luminescence import CommonLumi
 from lumispy.signals.common_transient import CommonTransient
 
 
+class TransientSpectrumCasting(Signal1D, CommonLumi, CommonTransient):
+    """**1D signal class for casting reduced LumiTransientSpectrum to either Luminescence or Transient**"""
```
ericpre, line 33: It would be good to explain in more details what it does, because it is an unusual pattern and it is not obvious what it does.

It would be good to add some example showing when it is used (even this will be hidden to the user) to help our future selves and other contributors.

lumispy/signals/luminescence_transientspec.py
```diff
@@ -21,21 +21,47 @@
...
+            "ps",
+            "ns",
+            "µs",
+            "mus",
+            "ms",
+            "s",
+        ]:
+            self.metadata.Signal.signal_type = "Transient"
+            self.__class__ = LumiTransient
+            self.__init__(*args, **kwargs)
```
ericpre, line 50: Simplify to:
```suggestion
            self.set_signal_type("Transient")
```

lumispy/signals/luminescence_transientspec.py
```diff
@@ -21,21 +21,47 @@
...
+        else:
+            self.metadata.Signal.signal_type = "Luminescence"
+            self.__class__ = LumiSpectrum
+            self.__init__(*args, **kwargs)
```
ericpre, line 54:
```suggestion
            self.set_signal_type("Luminescence")
```

ericpre (commented):
I don't like the idea of introducing the concept of "hidden signal" because by design registered signal class are not hidden.

If I understand correctly, the purpose of the `TransientSpectrumCasting` class is to assign to the correct class depending on the units of the last axis when converting to 1D signal. If so would it make sense to move this code to `LumiTransient.__init__` and change to the `Luminescence` signal type when necessary? It should work if the `TransientSpec` signal type was added as an alias of the `LumiTransient` class?

jlaehne:
@ericpre I was also hesitating to introduce a 'hidden signal class', but it works like a charm - while I cannot make it work without. 

It is more elegant than what I proposed in https://github.com/LumiSpy/lumispy/issues/204#issuecomment-1913154823 (and did not work), so I proposed this solution in https://github.com/LumiSpy/lumispy/issues/204#issuecomment-1913717244

I tried again having the initialization in the `__init__`  of one of the two existing 1D classes, but it then does not properly convert to the other. Instead, I have improved the documentation as you proposed.

jlaehne:
@ericpre do you have an idea why the changelog build displays a numba warning? Do I have to add that dependency to the workflow now that it is optional upstream?

Also, any idea on the Azure Pipelines failures?

ericpre:
> @ericpre do you have an idea why the changelog build displays a numba warning? See: https://lumispy--205.org.readthedocs.build/en/205/changelog.html Do I have to add that dependency to the workflow now that it is optional upstream?

This is surprising that it appears there! Yes, adding numba  as dependencies or using `hyperspy[speed]` can be used a workaround.

> Also, any idea on the Azure Pipelines failures?

mambaforge is deprecated, I will update what needs to be updated later today! :wink:

jlaehne:
@ericpre Do you want to have another look.

doc/user_guide/streak_images.rst
```diff
@@ -0,0 +1,53 @@
...
+
+.. code-block:: python
+
+    >>> s.sum(axis='Wavelength')
+    >>> # Or alternatively:
+    >>> s.sum(axis=-2)
+
+.. image:: images/streakmap.svg
+  :width: 700
+  :alt: Plot of a streak camera image and its one-dimensional representations
```
ericpre, line 52: Is there code in lumispy that can make this figure? When I do `lumispy.signals.luminescence_transientspec.LumiTransientSpectrum.plot`, I get a normal plot.

It would be good to add the code to make this figure, otherwise it should be removed?

lumispy/signals/common_luminescence.py
```diff
@@ -214,3 +224,300 @@ def normalize(self, pos=float("nan"), element_wise=False, inplace=False):
                 s.metadata.Signal.quantity = "Normalized intensity"
         if not inplace:
             return s
+
+    def _reset_variance_linear_model(self):
```
ericpre, line 228: Am I understanding correctly that this is simply moved around and doesn't need reviewed of the code itself?

upcoming_changes/205.new.rst
```diff
@@ -0,0 +1 @@
+Implement correct casting of :class:`~.signals.luminescence_transientspec.LumiTransientSpectrum` to either :class:`~.signals.luminescence_spectrum.LumiSpectrum` or :class:`~.signals.luminescence_transient.LumiTransient` when the signal dimension is reduced
```
ericpre, line 1:
```suggestion
Implement correct casting of :class:`~.signals.luminescence_transientspec.LumiTransientSpectrum` to either :class:`~.signals.luminescence_spectrum.LumiSpectrum` or :class:`~.signals.luminescence_transient.LumiTransient` when the signal dimension is reduced - previously, a :class:`~hyperspy.api.signals.Signal1D` object was returned.
```

ericpre (commented):
This looks good, I have some comments to improve the documentation. 

```python
import lumispy as lum
import numpy as np

data = np.arange(10*20).reshape(10, 20)

s = lum.signals.LumiTransientSpectrum(data)
s.axes_manager[-1].units = "ns"
s.axes_manager[-2].units = "nm"

s.sum(-1)
s.sum(-2)
```

lumispy/signals/common_luminescence.py
jlaehne: Yes, all the code pertaining to signal axis conversion is moved to the `CommonLuminescence` class so that it is available also to `LumiTransientSpectrum`. That was the origin of this PR and I probably should have merged that moving part in before doing the other changes to make it clearer.

doc/user_guide/streak_images.rst
jlaehne: I had previously prepared this figure for illustration purposes and thought it could be useful here to show the principle. It is indeed a useful feature to generally faciliate such kind of plots and I will open an issue to implement that in a separate PR (which could probably be taken on by one of my students). Not as the standard plotting of this class, but as a separate plot function (alternative to using `sum` an option to display spectra/transients for a certain range (or even an interactive span functionality can then be added later).

jlaehne:
> This looks good, I have some comments to improve the documentation.

I implemented your suggestions (and extended some to the rest of the documentation). Only the plot creation will need a separate PR.

doc/user_guide/streak_images.rst
```diff
@@ -27,25 +27,35 @@ assign the correct 1D signal class: Either :class:`~.signals.luminescence_spectr
...
 
 In the following image, the spectrum summed over all times is obtained from the
 streak image by:
 
 .. code-block:: python
 
-    >>> s.sum(axis='Time')
-    >>> # Or alternatively:
-    >>> s.sum(axis=-1)
+    import lumispy as lum
```
ericpre, line 38: Same as in https://github.com/LumiSpy/lumispy/pull/218#issuecomment-2436182282.

ericpre (approved):

PR #218, opened 2024-10-21
Adapt code display across user guide

jlaehne:
~Remove `>>>` from code displays to allow copying directly into notebooks.~

Adapt copybutton settings to remove `>>>` or similar formatting when copying code.

### Progress of the PR
- [x] ready for review.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,29 +21,29 @@ resources:
       # For more details on service connection endpoint, see
       # https://docs.microsoft.com/en-us/azure/devops/pipelines/library/service-endpoints
       endpoint: jordiferrero # Azure DevOps service connection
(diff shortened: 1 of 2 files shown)


ericpre:
How about using https://pypi.org/project/sphinx-copybutton instead of removing the prompt?
The advantage of keeping the prompt is to distinguish the input from the output.

jlaehne:
> How about using https://pypi.org/project/sphinx-copybutton instead of removing the prompt? The advantage of keeping the prompt is to distinguish the input from the output.

The latest development version is using `copybutton`: https://lumispy--205.org.readthedocs.build/en/205/user_guide/utilities.html
But somehow it does not remove the prompt when copying. Do I have to set some additional setting in `conf.py` for that?

PR #225, opened 2025-02-11
Add support for python 3.13

jlaehne:
- Add support for python 3.13
- Drop support for python 3.8
- Test for oldest supported hyperspy version and pin to >=v1.7.3

### Progress of the PR
- [x] Check formatting of the changelog entry (and eventual user guide changes) in the `docs/readthedocs.org:lumispy` build of this PR (link in github checks),
- [x] ready for review.
- [x] Test 3.13 against hs-RnMinor as https://github.com/hyperspy/hyperspy/pull/3468 is not yet released

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -86,3 +86,5 @@ jobs:
       - name: Upload coverage to Codecov
         if: ${{ always() }} && ${{ matrix.PYTEST_ARGS_COVERAGE }} 
         uses: codecov/codecov-action@v5
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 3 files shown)


ericpre (commented):
Would it be worth dropping python 3.8 support? It is not supported officially anymore: https://devguide.python.org/versions

jlaehne:
> Would it be worth dropping python 3.8 support? It is not supported officially anymore: https://devguide.python.org/versions

Valid point, I was wondering how far backwards to still support python versions when still supporting hyperspy 1.7.x

The multiple tests were just temporary for the moment to see which minor version of 1.7 ist still supported.

jlaehne:
@ericpre do you have a spontaneous idea why the 3.13 tests give an event trigger failure when a signal/axes object is part of a touple of possible test inputs?

ericpre:
Yes, it needs https://github.com/hyperspy/hyperspy/pull/3468 to be released!

PR #226, opened 2025-02-21
Release 0.3.0

jlaehne:
Release new version 0.3.0 following [releasing guide](https://github.com/LumiSpy/lumispy/blob/main/releasing_guide.md)

### Progress of the PR
- [x] Prepare release,
- [x] Changelog updated,
- [x] changelog verified,
- [x] test tag pushing in fork (see https://github.com/jlaehne/lumispy/actions/runs/13460214788/job/37613528129),
- [x] Released.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 Changelog entries for the development version are available at
 https://lumispy.readthedocs.io/en/latest/changelog.html
 
+UNRELEASED
+==========
 
 .. towncrier-draft-entries:: |release| [UNRELEASED]
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,38 @@ https://lumispy.readthedocs.io/en/latest/changelog.html
 
 .. towncrier release notes start
 
+0.3.0 (2025-02-21)
+==================
+
+New features
+------------
+
+- Implement correct casting of :class:`~.signals.luminescence_transientspec.LumiTransientSpectrum` to either :class:`~.signals.luminescence_spectrum.LumiSpectrum` or :class:`~.signals.luminescence_transient.LumiTransient` when the signal dimension is reduced - previously, a :class:`~hyperspy.api.signals.Signal1D` object was returned. (`#205 <https://github.com/lumispy/lumispy/issues/205>`_)
+- Add functions :meth:`~.signals.luminescence_transientspec.LumiTransientSpectrum.spec2nav` and :meth:`~.signals.luminescence_transientspec.LumiTransientSpectrum.time2nav` to convert either `signal` dimension to a `navigation` dimension for efficient iteration over the respective axes. (`#223 <https://github.com/lumispy/lumispy/issues/223>`_)
+
+
+Enhancements
+------------
+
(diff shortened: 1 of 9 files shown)


ericpre (approved):

jlaehne:
DOI assignment failed in the new setup that was copied from the other extensions.
@ericpre do you have some tips what might be configered wrongly? I did the linking at zenodo only after the release, but then edited the release and managed that zenodo tried to create a doi, but failed with `"errors": "401 Bad credentials"`.
@jordiferrero is there a kind of owner/administrator status at zenodo for the package that you could activate for me?

jlaehne:
Something went wrong on the version setting with `setuptools-scm`

After updating with conda to v0.3.0, it outputs:
```
>>> import lumispy as lum
>>> lum.__version__                                                             
'0.0.0'
```

ericpre:
Just tried `pip install lumispy` in a new environment and it works fine, I get the correct version (`0.3.0`).
I suspect that you are importing from a local source and not the version installed in `site-packages` because it seems to go through the [code path using `setuptools_scm`](https://github.com/LumiSpy/lumispy/blob/d6463676b5aad4a0e6b8e6b7a6b1117e0fea0856/lumispy/__init__.py#L44C14-L46).

To get the correct version when running from a version controlled source, you need the latest tag, because `setuptools_scm` uses the tags to build the dev version.

jlaehne:
I confirmed with a colleague having a clean conda environment and no dev versions on the PC that the conda install indeed prints `0.0.0` as version, when `0.3.0` is installed.

(the other part indeed is due to my existing local installation).

ericpre:
In my comment [above](https://github.com/LumiSpy/lumispy/pull/226#issuecomment-2677912820), this was with the package on pypi.org. With the conda-forge package, I can reproduce the issue and it is fixed in https://github.com/conda-forge/lumispy-feedstock/pull/9.

ericpre:
> DOI assignment failed in the new setup that was copied from the other extensions. @ericpre do you have some tips what might be configered wrongly? I did the linking at zenodo only after the release, but then edited the release and managed that zenodo tried to create a doi, but failed with `"errors": "401 Bad credentials"`.

Zenodo will create a DOI when a Github release is created (see for example: https://docs.github.com/en/repositories/archiving-a-github-repository/referencing-and-citing-content). Changing the workflow itself shouldn't affect this process.
Could it be this "bad credentials" error is related to an expired token or something similar.

> @jordiferrero is there a kind of owner/administrator status at zenodo for the package that you could activate for me?

I suspect the required permission to be able to do that are on github. Could it be the "third-party application access policy" setting at the organisation level?. On my zenodo account, I can't enable/disable the lumispy one (but in comparison, I can for exspy):
image

jlaehne:
> I suspect the required permission to be able to do that are on github. Could it be the "third-party application access policy" setting at the organisation level?. On my zenodo account, I can't enable/disable the lumispy one (but in comparison, I can for exspy):

Thanks @ericpre for your help with both issues. I was able to enable LumiSpy on Zenodo, but  I realize that @jordiferrero is the only one currently with permissions on the organization level to check the access policy.

jlaehne:
> In my comment [above](https://github.com/LumiSpy/lumispy/pull/226#issuecomment-2677912820), this was with the package on pypi.org. With the conda-forge package, I can reproduce the issue and it is fixed in [conda-forge/lumispy-feedstock#9](https://github.com/conda-forge/lumispy-feedstock/pull/9).

Thanks, it fixed the version issue.

PR #228, opened 2025-04-01
Add convoluted Gaussian and Exponential Component

jlaehne:
### Description of the change
Add a component that contains the convolution of an exponential decay with a Gaussian instrument response function.

### Progress of the PR
- [ ] update user guide (if appropriate),
- [x] added tests,
- [ ] ready for review.

```python
import lumispy as lum
m = s.create_model()
g1 = lum.components.ConvGaussianExponential()
m.append(g1)
m.fit()
```

### DIFF
--- a/lumispy/components/convolved_gaussian_exponential.py
+++ b/lumispy/components/convolved_gaussian_exponential.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2019-2023 The LumiSpy developers
+# Copyright 2019-2025 The LumiSpy developers
 #
 # This file is part of LumiSpy.
 #
@@ -16,16 +16,14 @@
 # You should have received a copy of the GNU General Public License
 # along with LumiSpy. If not, see <https://www.gnu.org/licenses/#GPL>.
 
-import math
-
 import numpy as np
+
 import hyperspy.api as hs
 
-sqrt2pi = math.sqrt(2 * math.pi)
-sigma2fwhm = 2 * math.sqrt(2 * math.log(2))
+sigma2fwhm = 2 * np.sqrt(2 * np.log(2))
 
 
-class ConvGaussianExponential(hs.model.components1D.Expression):
+class ConvGaussExp(hs.model.components1D.Expression):
     r"""Analytical convolution of Gaussian instrument response function and
     exponential decay function.
 
@@ -55,7 +53,7 @@ class ConvGaussianExponential(hs.model.components1D.Expression):
         Decay parameter (lifetime) of the exponential function.
     **kwargs
         Extra keyword arguments are passed to the
-        :py:class:`~._components.expression.Expression` component.
+        :py:class:`hyperspy._components.expression.Expression` component.
 
     Attributes
     ----------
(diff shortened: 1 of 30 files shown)


jlaehne:
@aidanc151 example of adding a component to lumispy :-)

lumispy/tests/components/test_convolved_gaussian_exponential.py
```diff
@@ -0,0 +1,70 @@
...
+sigma2fwhm = 2 * np.sqrt(2 * np.log(2))
+
+
+def test_function():
+    g = ConvGaussExp(height=3, t0=35, sigma=10, tau=100)
+    assert g.function(-50) == 0.0
+    np.testing.assert_allclose(g.function(35), 1.3874364)
+
+
+class TestConvGaussExp:
```
ericpre, line 36: Add a test that check that there is no normalisation issue: for example by making a dataset where the Gaussian convolution is negligible and checking that the fitting parameter are very close for the `ConvGaussExp` and `Exponential` components.

lumispy/components/convolved_gaussian_exponential.py
```diff
@@ -0,0 +1,109 @@
...
+        height=1.0,
+        t0=0.0,
+        sigma=10.0,
+        tau=100.0,
+        module=["numpy", "scipy"],
+        **kwargs,
+    ):
+        super().__init__(
+            expression="1/2*height*exp(-((x-t0) - sigma**2/(2*tau))/tau)*\
+                        (1 + erf(((x-t0) - sigma**2/tau)/(sqrt(2)*sigma)))",
```
ericpre, line 78: I didn't check the math but from a quick look (based on https://math.stackexchange.com/questions/685041/convolute-exponential-with-a-gaussian/872561#872561), it seems that it should be a "-"?
```suggestion
                        (1 - erf(((x-t0) - sigma**2/tau)/(sqrt(2)*sigma)))",
```
Could it be replaced by the `erfc` function?

lumispy/components/convolved_gaussian_exponential.py
```diff
@@ -0,0 +1,109 @@
...
+sigma2fwhm = 2 * np.sqrt(2 * np.log(2))
+
+
+class ConvGaussExp(hs.model.components1D.Expression):
+    r"""Analytical convolution of Gaussian instrument response function and
+    exponential decay function.
+
+    .. math::
+
+        f(x) = \frac{h}{2} \cdot \exp\left[(x - t_0 - \frac{\sigma^2}{2\tau})\tau^{-1}\right]
```
ericpre, line 32: Add a reference where the math comes from? 
For example, https://math.stackexchange.com/questions/685041/convolute-exponential-with-a-gaussian/872561#872561 has good pointers?

ericpre (commented):
It should be good to add this component to the document - I was looking for it to check for the rendering of the analytical expression!

lumispy/components/convolved_gaussian_exponential.py
jlaehne: So far, it is based on https://zenodo.org/records/6384277 - but we are checking the math again with mathematica. As perspective we will at least add a component with double exponential as well.

ericpre: Out of curiosity, are you using mathematica because sympy can't do it or just haven't tried with sympy?

jlaehne: A student already did the calculations in mathematica and I did not have the time yet to copy that in sympy :-)

jlaehne:
replaced by #262

PR #246, opened 2026-01-04
Lazy loading of modules and exposed utility functions

jlaehne:
### Description of the change
Import of LumiSpy is rather slow (`1.6 s` on linux) mainly due to the direct loading of `Signal1D` from HyperSpy. This PR uses <s>`__getattr__` as proposed by GitHub Copilot</s> `lazy-loader` to load both the sub-modules and utility functions exposed at the top level of LumiSpy using a lazy approach.

Splits off lazy signal modules in line with https://github.com/hyperspy/hyperspy/pull/3569.

### Progress of the PR
- [n/a] docstring updated (if appropriate),
- [n/a] update user guide (if appropriate),
- [X] added tests -> existing tests run through,
- [ ] ready for review.

### See also
Implementation in https://github.com/hyperspy/exspy/pull/160 and https://github.com/hyperspy/holospy/pull/65

```python
import time, importlib

t0 = time.perf_counter()
importlib.import_module("lumispy")
print("import time:", time.perf_counter()-t0)

```

### DIFF
--- a/lumispy/__init__.py
+++ b/lumispy/__init__.py
@@ -17,17 +17,14 @@
 # along with LumiSpy. If not, see <https://www.gnu.org/licenses/#GPL>.
 
 
+import importlib
+
 from importlib.metadata import version
 from pathlib import Path
-
-from lumispy.utils.axes import nm2eV, eV2nm, nm2invcm, invcm2nm, join_spectra
-from lumispy.utils.io import to_array, savetxt
-
-from lumispy import signals, components, utils
-
-from lumispy.utils import crop_edges
+from typing import Any
 
 
+# rresolve version
 __version__ = version("lumispy")
 
 # For development version, `setuptools_scm` will be used at build time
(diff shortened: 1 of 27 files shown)


ericpre (commented):
Is it worth adding test like in https://github.com/hyperspy/rosettasciio/blob/main/rsciio/tests/test_import.py?

Should [`lazy-loader` ](https://pypi.org/project/lazy-loader) be used to support typing and auto-completion in some development environment? This would add one more library but it is already a dependency of scikit-image and I also think that it will be good to use it in HyperSpy.

After https://github.com/hyperspy/hyperspy/pull/3569 is merged, it will be possible to speed up signals import even more to 0.2-0.3s.

ericpre:
```python
%time from lumispy.signals import CLSpectrum
```
takes 1.5 s on my laptop, while it should be possible to speed it up to ~0.3s.

ericpre (commented):
See https://github.com/jlaehne/lumispy/pull/20 for suggestions.

PR #255, opened 2026-02-20
implemented spectral response correction and calculate spectral response correction

Lulus792:
Implemented into CommonLumi the spectral_response_correction aswell as calculate_reponse_correction. Using a helper function _apply which takes an operator besides the ref signal since both function only differ in that functionality. Added tests aswell.

### Description of the change
- implemented spectral_response_correction
- implemented calculate_spectral_response

### Progress of the PR
- [x] added tests,
- [x] ready for review.

```python
import lumispy as lum
import numpy as np
ref = lum.signals.LumiSpectrum(0.2 * np.ones_like(np.arange(10)))
s.spectral_response_correction(ref, inplace=True)
```


ericpre:
How different is it from the following:
```python
import lumispy as lum
import numpy as np

ref = lum.signals.LumiSpectrum(0.2 * np.ones_like(np.arange(10)))
s = s * ref
```
If there is no difference, and if the https://peps.python.org/pep-0020/ is followed, `s = s * ref` should be preferred because it is simple and explicit and the method `spectral_response_correction` is not obvious what correction it would be applying.

upcoming_changes/255.new.rst
```diff
@@ -0,0 +1 @@
+Implemented spectral_reponse_correction aswell as calculate_spectral_response
```
jlaehne:
```suggestion
Implemented ``spectral_reponse_correction`` and ``calculate_spectral_response``
```

lumispy/signals/common_luminescence.py
```diff
@@ -233,6 +234,93 @@ def normalize(self, pos=float("nan"), element_wise=False, inplace=False):
...
+                        reference spectrum axis range.
+                        """)
+                sig = sig.isig[ref_min:ref_max]
+
+            ref = ref.interpolate_on_axis(sig_ax, -1, inplace=False)
+
+        result = op(sig, ref)
+        if inplace:
+            self.data = result.data
+            return self
```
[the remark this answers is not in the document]
Lulus792: the only reason I would return self is so that you can do chaining. Which is I believe usually what you do in programming that instead of void for a class you return the self so that you can perform chaining.

lumispy/signals/common_luminescence.py
```diff
@@ -233,6 +234,93 @@ def normalize(self, pos=float("nan"), element_wise=False, inplace=False):
...
+            ref_max = ref_ax.axis.max()
+            sig_min = sig_ax.axis.min()
+            sig_max = sig_ax.axis.max()
+            if sig_min < ref_min or sig_max > ref_max:
+                if sig_max < ref_min or sig_min > ref_max:
+                    raise ValueError("""
+                        The signal axis range does not overlap with the
+                        reference spectrum axis range.
+                        """)
+                sig = sig.isig[ref_min:ref_max]
```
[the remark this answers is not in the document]
Lulus792: I will aswell add a test to cover that functionality

lumispy/signals/common_luminescence.py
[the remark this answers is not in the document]
jlaehne: I would have thought that chaining in python works without the `return` statement, but I have actually not checked. Would be relevant for quite a few other functionalities in LumiSpy that do not have an explicit return for `self`. Could you check?

lumispy/signals/common_luminescence.py
```diff
@@ -233,6 +234,101 @@ def normalize(self, pos=float("nan"), element_wise=False, inplace=False):
...
+        Returns
+        -------
+        Signal object
+        """
+        return self._signal_math(ref, operator.mul, inplace)
+
+    def calculate_spectral_response(self, ref, inplace=False):
+        """Calculate the spectral response by dividing the signal by a reference
+        spectrum.
+
```
aidanc151, line 317: I think it might be worth adding more detail to the doc string, I'm not sure how obvious the context of this is to all users.  Maybe adding some examples?:

- a vendor provided reference spectrum of a calibration lamp, and your own spectrum of the lamp in your system
- a theoretical calculation of transition radiation for a reference, and your own spectrum of transition radiation in your system

If this is too much info for a doc string, then maybe putting it in the user guide and refering to the user guide in the docstring?

lumispy/signals/common_luminescence.py
```diff
@@ -233,6 +234,101 @@ def normalize(self, pos=float("nan"), element_wise=False, inplace=False):
...
+        result = op(sig, ref)
+
+        if inplace:
+            self.data = result.data
+            return self
+        return result
+
+    def spectral_response_correction(self, ref, inplace=False):
+        """Correct the spectral response of the signal by multiplying with a reference
+        spectrum.
```
aidanc151:
```suggestion
        spectrum. For example applying grating, detector or system spectral response curves.
```

aidanc151 (commented):
Looks good to me:) I tried it with some of my own spectra and also compared the `spectral_response_correction` to Domenik's personal function and they agree exactly

lumispy/signals/common_luminescence.py
aidanc151: by examples I don't mean to the Example gallery, but instead literal examples to the docstring

jlaehne: Docstring examples are only intended to showcase the syntax and use of parameters.

I would propose to create a user guide page (and possibly  an example in the gallery) once the code for calculating the transition radition response is also implemented. Just for this "wrapper" function it makes little sense. We could then also add a comment in this docstring that refers to the transition radiation function.

lumispy/signals/common_luminescence.py
jlaehne: @aidanc151 use "suggestion" feature for such simple corrections in the future :-)

upcoming_changes/255.new.rst
```diff
@@ -0,0 +1 @@
+Implemented ``spectral_response_correction`` and ``calculate_spectral_response``
```
jlaehne:
```suggestion
Implemented :meth:`~.signals.common_luminescence.spectral_response_correction` and :meth:`~.signals.common_luminescence.calculate_spectral_response`
```

jlaehne (approved):

upcoming_changes/255.new.rst
```diff
@@ -0,0 +1 @@
+Implemented :meth:`~.signals.common_luminescence.spectral_response_correction` and :meth:`~.signals.common_luminescence.calculate_spectral_response`
```
jlaehne:
```suggestion
Implemented :meth:`~.signals.common_luminescence.CommonLumi.spectral_response_correction` and :meth:`~.signals.common_luminescence.CommonLumi.calculate_spectral_response`
```

lumispy/signals/common_luminescence.py
```diff
@@ -233,6 +234,101 @@ def normalize(self, pos=float("nan"), element_wise=False, inplace=False):
...
+        if inplace:
+            self.data = result.data
+            return self
+        return result
+
+    def spectral_response_correction(self, ref, inplace=False):
+        """Correct the spectral response of the signal by multiplying with a reference
+        spectrum. For example applying grating, detector or system spectral response curves.
+
+        Patameters
```
jlaehne:
```suggestion
        Parameters
```

jlaehne:
I merge for now, we can include further documentation with the transient radiation functionality.

PR #260, opened 2026-05-13
spec2nav time2nav

janisschumacher:
Includes spec2nav and time2nav functions which return the streak image as signal with the time axis as navigation
axis and the spectral axis as signal axis and the other way around.

Interactive Version in which the intervals are selected interactively is also available.

Can be used like this:
streak01 = hs.load('../demo-files/CLEBIC2025/AlN_HR_streak_image.hspy')
streak22 = streak01.time2nav_tool(interactive = True, interval_count = 3)
streak22.result.plot()


jlaehne:
@ericpre, this function has an interactive version to select intervals to operate on in a GUI instead of passing e.g. a list of tuples. Do you see an easy implementation that the interactive function can return the same type of data as the non-interactive version? Currently, non-interactive directly returns a `newsignal`, but for interactive, the signal can accessed only by `newsignal.result`. Apparently, updating the new `newsignal` directly is the issue.

If I see it correctly, similar HyperSpy functionalities, such as `remove_background` or `spikes_removal_tool` operate inplace in the interactive version and only return a new signal in the non-interactive version.

PR #261, opened 2026-05-20
Speed up centroid function, only for uniform axis

Lulus792:
intergrated jordiferrero's faster centroid function, and split uniform and non uniform approach into two private functions, added a deprecationwarning to the old one.

### Description of the change
Left the centroid function for non-uniform unchanged. Only added a check for uniformity of the axis to the use jordiferreris's solution of PR 179.

### Progress of the PR
- [x] update user guide (if appropriate),
- [x] added tests,
- [x] Check formatting of the changelog entry (and eventual user guide changes) in the `docs/readthedocs.org:lumispy` build of this PR (link in github checks),
- [x] ready for review.

### DIFF
--- a/lumispy/signals/luminescence_spectrum.py
+++ b/lumispy/signals/luminescence_spectrum.py
@@ -21,6 +21,7 @@ Signal class for luminescence spectral data (1D).
 -------------------------------------------------
 """
 
+import lumispy
 import numpy as np
 from warnings import warn
 
@@ -78,7 +79,7 @@ class LumiSpectrum(Signal1D, CommonLumi):
         """
         warn(
             "The use of `remove_background_from_file` is deprecated and will "
-            "be removed in LumiSpy 1.0. Please use `remove_background_signal` "
+            "be removed in LumiSpy 1.0. Please use `remove_background` "
             "from the Signal1D class.",
             DeprecationWarning,
             stacklevel=2,
@@ -291,6 +292,7 @@ array([[ 0.,  0.,  1.,  2.,  3.,  4.],
         "\n", "\n\t"
     )
 
+    
     def centroid(self, signal_range=None, **kwargs):
         """Find the centroid (center of mass) of a peak.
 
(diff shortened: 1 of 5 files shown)


lumispy/signals/luminescence_spectrum.py
```diff
@@ -21,6 +21,7 @@
 -------------------------------------------------
 """
 
+import lumispy
```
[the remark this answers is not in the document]
jlaehne: The later `import from` statements do the job, so this one should not be necessary.

lumispy/utils/signals.py
```diff
@@ -97,15 +97,21 @@ def com(spectrum_intensities, signal_axis, **kwargs):
...
+        "in LumiSpy 1.0. Please use `lumispy.signals.luminescense_spectrum.centroid` instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    
     def _interpolate_signal(axis_array, index, **kwargs):
         """Wrapper for `hs.axes.index2value`.
-
-        Linearly interpolate between values should the index passed not be a
+
```
jlaehne, line 109: unnecessary whitespace (weird that lint does not complain)

lumispy/utils/signals.py
```diff
@@ -97,15 +97,21 @@ def com(spectrum_intensities, signal_axis, **kwargs):
...
         """Wrapper for `hs.axes.index2value`.
-
-        Linearly interpolate between values should the index passed not be a
+        
+        Linearly interpolate between values should the index passed not be an
         integer. Using the kwargs, the interpolation method can be changed.
         """
         rem = index % 1
-        index = int(index // 1)
+        index  = int(index // 1)
```
jlaehne, line 114: unnecessary whitespace (weird that lint does not complain)

jlaehne:
Looks good. Can you add a set of tests based on real data using the summed up spectrum from `asymmetric_peak_map`:
a) directly for the data (uniform axis)
b) do energy conversion and then determine centroid (nonuniform axis).

If you determine the result values with the old implementation, this test will directly assert that the behaviour is not changed by this PR.

Adapt the tolerance of `assert`, as it might be prone to failing if it is the strict default value.

lumispy/utils/signals.py
```diff
@@ -97,15 +97,21 @@ def com(spectrum_intensities, signal_axis, **kwargs):
     >>> center_of_mass = com(intensities, signal_axis)
     >>> print(center_of_mass)  # Outputs: [400.0]
     """
-
+    warn(
+        "The use of `lumispy.utils.signals.com` is deprecated and will be removed "
+        "in LumiSpy 1.0. Please use `lumispy.signals.luminescense_spectrum.centroid` instead.",
```
jlaehne:
```suggestion
        "in LumiSpy 1.0. Please use the object-oriented `centroid` function of the `luminescense_spectrum` class instead.",
```

lumispy/utils/signals.py
```diff
@@ -97,15 +97,21 @@ def com(spectrum_intensities, signal_axis, **kwargs):
...
+        stacklevel=2,
+    )
+    
     def _interpolate_signal(axis_array, index, **kwargs):
         """Wrapper for `hs.axes.index2value`.
-
-        Linearly interpolate between values should the index passed not be a
+        
+        Linearly interpolate between values should the index passed not be an
         integer. Using the kwargs, the interpolation method can be changed.
```
jlaehne:
```suggestion

       If the index passed is not an integer,  linearly interpolate between values.
       Using the kwargs, the interpolation method can be changed.
```

jlaehne (approved):

lumispy/utils/signals.py
jlaehne:
```suggestion
        index = int(index // 1)
```

PR #263, opened 2026-06-24
Implementing transition radiation modelling

aidanc151:
### Description of the change
In cathodoluminescence, experimentally recorded transition radiation (TR) spectra can be compared to theoretically modeled TR - as an approach to obtain the spectral response of optical setup.  DOI10.1063/1.4885426

This PR adds a utility function for executing Toon Coenen's .py calculations for TR modelling. Optical constants are required for the calculation therefore a folder of curves (from literature) is also added. 

Currently the function rather lazily just executes Toon's .py, in the future the calculations can be integrated into lumispy more appropriately and efficiently.

### Issues/to be resolved
Currently the function outputs both a `Signal1D` spectrum and a `Signal2D` angle-resolved spectrum. The latter - though interesting, is maybe quite specialised. Additionally, this data is usually inspected in a polar plot (which I don't think hyperspy supports) and I would also wonder if a matplotlib plotting example should be included in that case. I'm leaning more towards removing the angle-resolved spectrum output, and instead mentioning in the documentation that angle-resolved data can also be calculated using Toon's .py directly. 

### Progress of the PR
- [ ] resolve function outputs (described above)
- [ ] function naming?
- [ ] determine placement of optical constant files and TR calculation file 
- [ ] make optical constant options known in docstring 
- [ ] update user guide
- [ ] added tests
- [ ] ready for review.

```python
spec, ang  = lum.utils.axes.transition_radiation_nm(
    electron_energy=30,
    material_dataset='Al_mcpeak',
    angular_axis=(0, 90),
    spectral_axis=(110, 1200),
)

spec.plot()
```


jlaehne:
FYI, I fixed the Release&Package workflow permissions and added failing links to `linkcheck_ignore` in the upstream `conf.py` so that tests should not be failing in the future.

lumispy/data/transition_radiation/optical_constants/Ag_johnson.csv
```diff
@@ -0,0 +1,50 @@
+wl,n,wl,k
```
jlaehne: I would say `transition_radiation/optical_constants` is one level too much in the folder structure, how about just going for a folder `optical_constants` in `data`.

I would consistently convert to one data format, e.g. `.csv` with commas (not tabs) and first line as index. It is good to be more flexible in reading, but with the preshipped data I would be consistent. Alternative would be to actually use HyperSpy data files that would allow adding citation data to the metadata (if n and k are given there would need to be a navigation dimension).

Add a `Readme.md` file that concisely documents the data format? The option to read in `csv/txt` files could stay even if we go for HyperSpy files to make it easier for people to provide their own data without converting - but in principle HyperSpy can read text files and the only question is how it handles e.g. four column files with two spectral axes - but if we require and document a `wl,n,k` style in that case should simply work with hs.read(`optical.csv`)?

